PR #4, opened 2020-11-16
Upgrade Rollbar maybe

dajinchu:
bakcend server is not talking to rollbar properly, particularly on uncaught exceptions. trying to use the rollbar's own `captureUncaught` instead of our own wack stuff in hopes that it'll help. Also upgraded rollbar version


gamburgm (approved):
:+1:

PR #5, opened 2020-11-16
Bug When Search Course No Sections

megandouli:
# What
Server crashes when people do a single search result for a course with no sections

# Why
To get aggregation information for a single search result, we assumed there'd be at least one section and tried to get the class type of the first section.

# Fix
Added a check in method that gets aggregation: if there's no section for that course, the classType aggregation is an empty array

# Questions
Should a single search result even show up if there are no sections? Right now, if we search "ARTG 5100" in the Spring, we see the course information but no sections (looks a bit odd).

Screen Shot 2020-11-16 at 3 01 45 PM

### DIFF
--- a/backend/searcher.ts
+++ b/backend/searcher.ts
@@ -2,7 +2,6 @@
  * This file is part of Search NEU and licensed under AGPL3.
  * See the license file in the root folder for details.
  */
-
 import _ from 'lodash';
 import { Course, Section } from '@prisma/client';
 import prisma from './prisma';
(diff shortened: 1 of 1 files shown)


backend/searcher.ts
```diff
@@ -262,7 +262,7 @@ class Searcher {
     return {
       nupath: result.nupath.map((val) => { return { value: val, count: 1 } }),
       subject: [{ value: result.subject, count: 1 }],
-      classType: [{ value: result.sections[0].classType, count: 1 }],
+      classType: result.sections[0] ? [{ value: result.sections[0].classType, count: 1 }] : [],
```
NEUDitao: nice

NEUDitao (approved):

PR #8, opened 2020-11-18
Ml single search result bug

megandouli:
# what
Crashes when people make a single search result for a course with no sections

# why
Made an assumption that if the course exists, there must be sections

# fix
Added a check to see if a course exists AND has sections, if not, we return no search results


NEUDitao (approved):

gamburgm (approved):
:+1:

PR #2, opened 2020-11-15
Campus filter

ghost:
# Why 

So that people can more finely search for courses by their location. Alex showerthought: this might be better named "location filter"... is MFA a campus?

# What 

Add a filter for locations. Ie, only show results with the selected locations.

# Spec

Figma Mockup: https://www.figma.com/file/Eh1nMa7vQJ0mxSNIHvyI51/%E2%9C%85-Dev-Ready?node-id=305%3A55

Trello Card: https://trello.com/c/Kd6CfNp5/58-campus-filter

Picture of the Figma mock:

Screen Shot 2020-11-12 at 10 25 44 PM

# Todos

- [x] update db models
- [x] update scrapers
- [x] update backend filters and types
- [x] update frontend filters and types

Daniel's diff: https://github.com/sandboxnu/searchneu/pull/176/files/98cf72f834e2ba042004a280acb976b9bd32278f

PR #23, opened 2020-11-29
Campus filter (rebased on master)

ghost:
# Why

I had a decent bit of trouble rebasing onto master, and I wanted to make sure that my changes still function properly, so this PR is the same filter code just rebased onto master.

# What

The old PR (#2) rebased onto master + more recent master merge + fix for single search result bug.

# Instructions to running code

1. `git checkout campusFilter2`
2. `yarn install` (or just yarn, thanks @zefwang )
3. `yarn db:refresh`
4. `yarn db:migrate` - this might fail with the message `Local and remote migrations are not in lockstep...` in which you need to drop your database, see later instructions
5. `yarn scrape`
6. `yarn start`

### Instructions to fixing the DB problem

You can either drop your database, or run `yarn db:rollback` until it works. These instructions aren't very specific (i'm sort of hoping that you don't have to do this). If so, lmk and i can send over more instrs.

link to old PR: #2


megandouli (approved):
looks good! The scrapers don't work but that's not on your end. I did a hack-y fix locally to get the scrapers working and then everything with campus filters works fine!

ghost:
thanks megan :)

PR #26, opened 2020-12-02
CPS Homepage, Mobile, Results Dropdown

ghost:
### What
built off of Amiel's work to fix some small styling on the home page search + add in the breadcrumbs dropdown to the results page for campus + term

### Why
This is a part of the cps/law integration

### Notes
- I did not add a transition when changing schools on the home page (I initially thought we needed it but I think it looks fine)
- If you click to go back to the home page the semester/term that was chosen is not synced, should it be?

### Tickets
- https://trello.com/c/i2gl27e3/101-displaying-schools-on-home-screen
- https://trello.com/c/37caHWuo/121-mobile-designs-for-cps-law
- https://trello.com/c/aXWfjdSs/102-displaying-schools-on-home-search-results-page


ghost (commented):
@epillay note from @beccajohnson17 and @puneetjohal -- we need to fix the option duplication still on the Homepage.

also I noticed that if the options get wide, they still look weird :( in the homepage dropdown

beccajohnson17 (changes requested):
As discussed on the slack thread (putting here for easier accessibility):

- Adding indication between cps and law semesters (similar to Banner, they have all the semesters for all schools under one big dropdown so they need to specify a bit more)

- "Spring Semester 2020" shows up at the top of the dropdown when changing schools or if the term is changed and keeps duplicating.

- Not seeing results (could be specific to me, sometimes I see no data and other times I get a white screen)

megandouli (commented):
I just ran it on my machine and it is BEAUTIFUL!!!

puneetjohal (changes requested):
There is no data for the spring 2021 quarter for cps or law (and it's not up on banner yet either), so I think we should get rid of those from the homepage/breadcrumb dropdowns, since they just lead to a no results page.

puneetjohal (approved):
chef's kiss ✨

megandouli (approved):
Everything looks good to me! I like the CPS/Law indication and I don't see duplicate spring semesters :)

beccajohnson17 (approved):

beccajohnson17:
BEAUTIFUL!

PR #27, opened 2020-12-03
New Homepage

ghost:
### todo

- [x] add in sandbox logo and link
- [x] change law color from green -> blue
- [x] make dropdown appear in front of buildings
- [x] resize logo in result page
- [ ] make sure mobile looks good 
- [x] change color based on the state (campus)
- [x] change magnifying glass


epillay:
## todo
- [ ] fix mobile dropdown for long names

epillay:
we still have to fix long names in the dropdown for mobile, but this will be in a separate PR so that the rest of the changes can be merged in!

frontend/components/ResultsPage/SearchBar.tsx
```diff
@@ -4,19 +4,20 @@
...
+import magnifyingGlass from '../images/magnifying-glass.svg';
 
 interface SearchBarProps {
-  query: string,
-  onSearch: (q: string) => void,
-  onClick?: () => void,
+  query: string;
+  onSearch: (q: string) => void;
+  onClick?: () => void;
+  buttonColor: 'red' | 'yellow' | 'blue';
```
megandouli, line 13: The `buttonColor` type seems duplicated in HomeSearch.txt. Just wondering, should this be its own type or would that be overkill?

megandouli (changes requested):
Amazing amazing amazing!!! Can't believe how quickly you guys got this done

zefwang (commented):
Looks great :)

frontend/components/ResultsPage/SearchBar.tsx
epillay: I think because this is so specific to the homepage that for now, it is okay. But I think if we need to expand the color/state relationship to the splash pages and results screen then it should definitely be a type

megandouli (approved):
❤️

beccajohnson17 (changes requested):
THIS IS SO HYPE GUYS IT LOOKS AMAZING!!!

- Currently, the dropdown for the search bar disappears behind the skyline. I tried exploring it a bit in dev tools and changing the z-index on .HomeSearch__searchBar--dropdown seemed to resolve the issue, but there might be a more efficient way.

- *not something to worry about before putting this into prod but something we should do soon: Making the husky responsive. Right now, it eventually overlaps over the search bar when you reduce the screen width. See pic for reference.

Screen Shot 2020-12-04 at 11 10 30 PM

beccajohnson17:
just to reiterate, this is absolutely beautiful!!!

beccajohnson17 (approved):

PR #28, opened 2020-12-04
Single Search Result Subject Description

megandouli:
# what
Single search result doesn't show subject description

# why
Aggregation generation is different for single search result compared to normal searches

# fix
Called helper function for aggregation generation for subject filter


gamburgm (approved):
:+1:

PR #31, opened 2020-12-07
Bug Rendering Prereqs

megandouli:
# what
Prereqs were improperly displayed. CS4500 should have prereqs: (ENGW 1102 or ENGW 1111) and CS 3500
but results were  (ENGW 1102 and ENGW 1111) and CS 3500

# fix
The backend was fine, the frontend just didn't render the type of nested prereqs correctly.


gamburgm (approved):
:+1: I missed this when looking at the module, good fix!

PR #33, opened 2020-12-08
Remove Staff from Results

gamburgm:
Or, why the f*** are we showing electricians as search results?


lizzij (approved):
LGTM

puneetjohal (approved):
Very complicated

PR #42, opened 2020-12-13
dc/graphql search

dajinchu:
## Overview
This PR adds search to graphql. You can preview using the playground in browser at `/graphql`

**Note that the return value is a little different from REST.**
We used to send back a `type` field in each result item. GraphQL convention calls it `__typename`. Potato potato, same thing.
Also note that REST search items looked like 
```json
{
  "type": "class",
  "class": {},
  "sections": {}
}
```
Whereas the graphql structure is 
```json
{
   "__typename": "ClassOccurrence",
  "name": "Fundies",
  "subject": "CS",
  "sections": {}
}
```
with the section held as just a property of the class occurrence. I think this just makes more sense logically. ClassOccurences should have references to their sections. Simple.

## Deploying

This PR is low risk. Just adds a graphql query. Does not impact existing functionality.

## Example Query 
```graphql
{
  search(termId: 202130, 
    subject:"CS", 
    classIdRange:{max:5000,min:4000},  
    first: 2){
    totalCount
    pageInfo {
      hasNextPage
    }
    nodes {
      __typename
      ... on Employee {
        name
      }
      ... on ClassOccurrence {
        name
        sections {
          profs
          crn
        }
      }
    }
  }
}
```

**Result**
```json
{
  "data": {
    "search": {
      "totalCount": 15,
      "pageInfo": {
        "hasNextPage": true
      },
      "nodes": [
        {
          "__typename": "ClassOccurrence",
          "name": "Artificial Intelligence",
          "sections": [
            {
              "profs": [
                "Lok Sang Wong"
              ],
              "crn": "30209"
            },
            {
              "profs": [
                "Stacy Marsella"
              ],
              "crn": "38417"
            }
          ]
        },
        {
          "__typename": "ClassOccurrence",
          "name": "Natural Language Processing",
          "sections": [
            {
              "profs": [
                "Felix Muzny"
              ],
              "crn": "32493"
            }
          ]
        }
      ]
    }
  }
}
```

### DIFF
--- a/backend/graphql/index.js
+++ b/backend/graphql/index.js
@@ -2,6 +2,11 @@ import { ApolloServer, gql } from 'apollo-server-express';
 import GraphQLJSON, { GraphQLJSONObject } from 'graphql-type-json';
 import macros from '../macros';
 
+import employeeTypeDef from './typeDefs/employee';
+
+import searchResolvers from './resolvers/search';
+import searchTypeDef from './typeDefs/search';
+
 import classResolvers from './resolvers/class';
 import classTypeDef from './typeDefs/class';
 import classOccurrenceTypeDef from './typeDefs/classOccurrence';
@@ -9,7 +14,6 @@ import classOccurrenceTypeDef from './typeDefs/classOccurrence';
 import majorResolvers from './resolvers/major';
 import majorTypeDef from './typeDefs/major';
 import majorOccurrenceTypeDef from './typeDefs/majorOccurrence';
-
 
 // Enable JSON custom type
 const JSONResolvers = {
@@ -28,8 +32,16 @@ const baseQuery = gql`
 `;
 
 const server = new ApolloServer({
-  typeDefs: [baseQuery, classTypeDef, classOccurrenceTypeDef, majorTypeDef, majorOccurrenceTypeDef],
-  resolvers: [JSONResolvers, classResolvers, majorResolvers],
+  typeDefs: [
+    baseQuery,
+    classTypeDef,
+    classOccurrenceTypeDef,
+    employeeTypeDef,
+    majorTypeDef,
+    majorOccurrenceTypeDef,
+    searchTypeDef,
+  ],
+  resolvers: [JSONResolvers, classResolvers, majorResolvers, searchResolvers],
 });
 
 if (require.main === module) {
--- a/backend/graphql/index.js
+++ b/backend/graphql/index.js
@@ -29,6 +29,11 @@ const baseQuery = gql`
   type Query {
     _empty: String
   }
+
+  type PageInfo {
+    # When paginating forwards, are there more items
+    hasNextPage: Boolean!
+  }
 `;
 
 const server = new ApolloServer({
--- a/backend/graphql/resolvers/search.ts
+++ b/backend/graphql/resolvers/search.ts
@@ -1,7 +1,52 @@
+import { identity, pickBy, result } from 'lodash';
+import searcher from '../../searcher';
+import { SearchResult } from '../../search_types';
+import { Course, Employee } from '../../types';
+
+interface SearchArgs {
+  termId: number;
+  query?: string;
+  subject?: string[];
+  nupath?: string[];
+  campus?: string[];
+  classType?: string[];
+  classIdRange?: { min: number; max: number };
+}
 const resolvers = {
   Query: {
-    search: (parent, args) => {
-      return {};
+    search: async (
+      parent,
+      args: SearchArgs,
+    ): Promise<(Course | Employee)[]> => {
+      const results = await searcher.search(
+        args.query || '',
+        String(args.termId),
+        0,
+        10,
+        pickBy(
+          {
+            subject: args.subject,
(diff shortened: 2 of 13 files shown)


backend/graphql/index.js
```diff
@@ -25,11 +29,24 @@ const baseQuery = gql`
   type Query {
     _empty: String
   }
+
+  type PageInfo {
```
gamburgm, line 33: Don't we only have this property on the search endpoint (and if so, shouldn't this just be a property there)?

backend/graphql/resolvers/search.ts
```diff
@@ -0,0 +1,73 @@
+import { identity, pickBy } from 'lodash';
+import searcher from '../../searcher';
+import { Course, Employee } from '../../types';
+
+type SearchResultItem = Course | Employee;
```
gamburgm, line 5: Nit: `SearchResult` vs. `SearchResultItem`, `SearchResultMeta` vs. `SearchResultItemConnection`? I care less about the former than the latter, I don't think `connection` accurately describes the data here

backend/graphql/resolvers/search.ts
```diff
@@ -0,0 +1,73 @@
...
+  campus?: string[];
+  classType?: string[];
+  classIdRange?: { min: number; max: number };
+
+  offset: number;
+  first: number;
+}
+const resolvers = {
+  Query: {
+    search: async (
```
gamburgm: Nit but can you put the function declaration on one line?

backend/graphql/resolvers/search.ts
```diff
@@ -0,0 +1,73 @@
...
+  termId: number;
+  query?: string;
+  subject?: string[];
+  nupath?: string[];
+  campus?: string[];
+  classType?: string[];
+  classIdRange?: { min: number; max: number };
+
+  offset: number;
+  first: number;
```
gamburgm, line 25: What is `first` supposed to mean? Is this supposed to be like `first ten` kinda thing? If so, then I think `count` or `size` would be better

backend/graphql/resolvers/search.ts
```diff
@@ -0,0 +1,73 @@
...
+}
+const resolvers = {
+  Query: {
+    search: async (
+      parent,
+      args: SearchArgs,
+    ): Promise<SearchResultItemConnection> => {
+      const { offset = 0, first = 10 } = args;
+      const results = await searcher.search(
```
gamburgm, line 31: In theory you could destructure here for `resultCount` and `searchContent`, but no big deal either way.

backend/graphql/typeDefs/classOccurrence.js
```diff
@@ -7,11 +7,35 @@ const typeDef = gql`
...
 
-    desc: String
+    desc: String!
     prereqs: JSON
     coreqs: JSON
+    prereqsFor: JSON
+    optPrereqsFor: JSON
     maxCredits: Int
     minCredits: Int
+    classAttributes: [String!]!
```
gamburgm, line 17: I think it's possible for `classAttributes` to be empty, how would you express that?

backend/search_types.ts
```diff
@@ -208,21 +206,14 @@ export interface AggResults {
...
-export interface SearchResult {
-  [resultId: string]: ResultDoc;
-}
-
-export interface CourseDoc {
-  class: Course;
-  sections: Section[];
-}
-export type ResultDoc = Employee | CourseDoc;
+export type SearchResult =
```
gamburgm, line 214: This is not directly related to this ticket but should we change the model to not have these `type` declarations?

gamburgm (changes requested):
Looks pretty good but added some comments. Also, the type errors in `updater.test.seq.ts` I think are caused by this PR (we'll get rid of the `dumpProcessor` type errors in the re-write). 

What's the plan for typescript GraphQL?

backend/graphql/index.js
dajinchu: It's just a GraphQL convention for pagination. We will likely need it in many places.

backend/graphql/resolvers/search.ts
dajinchu: https://graphql.org/learn/pagination/#pagination-and-edges Just GraphQL convention.

backend/graphql/resolvers/search.ts
dajinchu: GraphQL convention uses Item and Connection. Some graphql clients like Relay require it to be in this format to auto-handle pagination. 

For example: https://docs.github.com/en/free-pro-team@latest/graphql/reference/queries#searchresultitemconnection

backend/graphql/typeDefs/classOccurrence.js
dajinchu: Empty is allowed by this type def. `[Type]!` means it must be an array and not null. `[Type!]` means the items must be of Type and not null.
`[]` is an array with items that are not null.

NEUDitao (approved):

gamburgm (approved):
:+1:

PR #53, opened 2020-12-26
Get Scrapers Working

gamburgm:
# What
Scrapers fail on prod and for a number of local machines, so want to get it _passing_ (virtually no matter the cost--we'll add back features and clean up code later).

# How
1. Improving performance by doubling the number of open connections and having the number of simultaneous requests equal to the number of possible open connections (potentially reduce thrashing).
2. Don't search for classes referenced in `reqs`. It seemed to be the root cause of the failures.

# Notes
1. Want to bring the deleted feature back later.
2. We need a new request system that at least logs more expressively. Our request infrastructure may also need a revamp. The code is quite bad.

### DIFF
--- a/backend/scrapers/classes/parsersxe/termParser.js
+++ b/backend/scrapers/classes/parsersxe/termParser.js
@@ -25,6 +25,7 @@ class TermParser {
     const subjectTable = await getSubjectDescriptions(termId);
     const sections = await this.parseSections(termId);
     const courseIdentifiers = {};
+
     sections.forEach((section) => {
       const subject = section.subject;
       const classId = section.classId;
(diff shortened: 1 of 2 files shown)


megandouli (approved):
Nice! Scraping 202130 works for me! Do you want the scrapers to only scrape that one term?

hankewyczz:
> I put this PR up back when the scrapers were very flakey, and had different failure rates on each machine (iirc) — it consistently worked on Megan’s but rarely on mine
I was trying to figure out what was going on just to quick fix with the plan of properly repairing it later, I guess the observation I made was that changing the concurrency count increased reliability
In any case, the PR is almost certainly useless and out of date, so you can safely close

PR #56, opened 2020-12-29
NextJS part 1!!!

### LINKED ISSUE #49: Next.js
# Goal
Next.js means a few things:
1. We can make SearchNEU serverless
2. Code organization is much cleaner, and we can delete a bunch of crap (e.g. React Router)
3. We can speed up load times with server rendering and code splitting

# Design
1. Re-structure the frontend to use Next.js, but leave current backend alone
2. Add Vercel integration
3. post-split, add a route for the notification workflow code

Once the above is finished, we can toss the backend code (in theory).

NEUDitao:
So we're not *actually* merging this, but, this is just a PR for people to see the *first* part of NextJS. This was just a migration to get the app working in NextJS, and delete the backend completely. The intent of this is not to merge, but rather to allow people to approve before we keep merging into this branch.

Oh god I hope my line count is restored after this.

To run, all you have to do is checkout the branch, `yarn`, and then `yarn dev`.

Warning! Don't know if it breaks compatibility with the main app, but *shouldn't*. Just keep careful track of all of the untracked files and you should probably be good. 

Closes #49


components/ResultsPage/Results/useUserChange.ts
```diff
@@ -1,10 +1,11 @@
-import { useState, useEffect } from 'react'
-import Course from '../../classModels/Course'
-import user from '../../user'
-import Keys from '../../../../common/Keys'
+import { useEffect, useState } from 'react';
```
NEUDitao: This is part of the "fix `window` is not defined" suite -- just puts the `user.isWatchingClassHAsh` in the `useEffect` so it's only done on the client

pages/404.tsx
```diff
@@ -0,0 +1,3 @@
+export default function NotFoundPage() {
+    return <h1>YOU GOT LIGMA</h1>;
```
NEUDitao: we *should* have a real 404 page at some point

components/search.ts
```diff
@@ -77,7 +77,7 @@ class Search {
     macros.log('Requesting terms ', existingTermCount, 'to', termCount);
 
 
-    const url = new URI('/search').query({
+    const url = new URI('https://searchneu.com/search').query({ // TODO: is this how we're gonna access the api in the future?
```
NEUDitao: is this how we're gonna access the api in the future? or are we hosting it somewhere else. this does also raise the question -- do we know if anyone else uses our API?

dajinchu:
It'll be a bit easier to review this using [this link](https://github.com/sandboxnu/searchneu/pull/56/files/1cdf6a4fd270f1d1b1d1a09b5e5c6c065f4db77c..4af578c684e40ceea045afdbd59471366e70ee44) which ignores the first commit which just moved/deleted files

dajinchu (approved):

dajinchu:
Great PR, good job. Such beautiful code good wokr

PR #67, opened 2021-01-04
New Major API

gamburgm:
Following the new spec, which is attached below:

```graphql
type MajorQuery extends Query {
  major(majorId: String!): Major
}

type Major {
  majorId: String!

  occurrence(year: Int!): MajorOccurrence!
  latestOccurrence: MajorOccurrence!
}

type MajorOccurrence {
  majorId: String!
  yearVersion: String!

  spec: JSON!
  plansOfStudy: JSON!
}
```


backend/graphql/tests/__snapshots__/major.test.seq.ts.snap
```diff
@@ -2,11 +2,9 @@
 
 exports[`cannot find major from non-present majorId 1`] = `
 Object {
-  "data": Object {
-    "major": null,
-  },
+  "data": undefined,
   "errors": Array [
-    [GraphQLError: major not found!],
+    [ValidationError: Cannot query field "majorId" on type "Query". Did you mean "major"?],
```
megandouli: Is this ValidationError supposed to be in the snapshot?

backend/graphql/tests/__snapshots__/major.test.seq.ts.snap
```diff
@@ -19,14 +17,9 @@ Object {
...
-  "data": Object {
-    "major": Object {
-      "name": "Computer Science",
-      "occurrence": null,
-    },
-  },
+  "data": undefined,
   "errors": Array [
-    [GraphQLError: occurrence not found!],
+    [ValidationError: Cannot query field "majorId" on type "Query". Did you mean "major"?],
```
megandouli: ^^ Same as above

scripts/migrate_major_data.ts
```diff
@@ -1,40 +1,51 @@
...
+      }
     });
-  });
+  }, { concurrency: CONCURRENCY_COUNT });
 }
 
-migrateData(fetchData('major.json') as MajorJSON);
+(async () => {
+  await migrateData(fetchData().all_objects);
+  console.log('Success! You may close.');
```
megandouli: Do we want to also log the amount of time it took to migrate all the data?

megandouli (approved):
Overall looks good! Just want to confirm the test snapshot is what we want

dajinchu (approved):

backend/graphql/tests/__snapshots__/major.test.seq.ts.snap
gamburgm: WOW, great catch, just fixed this.

scripts/migrate_major_data.ts
gamburgm: Great idea!

megandouli (approved):
chungus

PR #68, opened 2021-01-08
Homepage edits

ghost:
We made some edits to the Homepage based on feedback from Diana.

https://www.notion.so/SearchNEU-re-design-edits-75d173a695234bd5be906559b4ed9bf8

Waiting for Diana to validate before merging, but also probably should get code review.


NEUDitao (approved):
code looks clean to me

PR #73, opened 2021-02-05
Database Port Change

megandouli:
Changed search database to run on port 5433 instead of 5432 to avoid conflicts with course-catalog-api. Added `db:refresh` command to package.json and renamed` migrate` to `db:migrate`


puneetjohal (approved):
🐸

PR #75, opened 2021-02-07
Update tie icon to be the correct shade of yellow

puneetjohal:
The CPS tie icon in the modal on the home page was the wrong shade of yellow, but is now the correct one.


megandouli (approved):
Looks good! Merging it in

PR #76, opened 2021-02-08
Update term names for CPS and Law to no longer include the school name

puneetjohal:
The textual term names for CPS and Law include the school names in them, making our term dropdown options look redundant. This change removes the school names from the term mappings in [`./components/global.ts`](https://github.com/sandboxnu/searchneu/blob/pj-update-cps-law-term-names/components/global.ts#L14-L38).

Now instead of saying something like `Winter 2020 Law Quarter`, the term dropdown will instead say `Winter 2020 Quarter`. It's already clear which school we're referring to based on the states of other components on the home/results page.


megandouli (approved):
Merging this into master :)

PR #83, opened 2021-02-21
Convert to GraphQl

megandouli:
## What
Converting search to use the GraphQL endpoint instead of Rest. 
See https://github.com/sandboxnu/course-catalog-api/pull/11 as well.

## How
Hit the GraphQL endpoint and transform the results we get back to mirror the shape of the old rest API results.

## Steps To Test
- Make sure `course-catalog-api` is running locally (if the branch `ml-hit-graphql` in `course-catalog-api` hasn't been merged into master yet, switch to that branch)
- In your local search repo, go to the file `codegen.yml` and replace `schema: 'https://searchneu.com/graphql'` with `schema: 'http://localhost:4000'`
- In your local search repo, go to `utils/courseAPIClient.ts` and replace `const ENDPOINT = 'https://searchneu.com/graphql';` with `const ENDPOINT = 'http://localhost:4000'`


NEUDitao (approved):

NEUDitao (commented):
can you pull from master so we have running unit tests (they're gonna fail on `notify_users` but that's no biggy), and then also fix up the typescript errors?

components/ResultsPage/useSearch.ts
```diff
@@ -78,7 +83,19 @@ export default function useSearch({
   const { data, size, setSize } = useSWRInfinite(
     getKey,
     async (query): Promise<SearchResult> => {
-      return (await axios.get('https://searchneu.com/search?' + query)).data;
+      const urlParams = Object.fromEntries(new URLSearchParams(query));
+      const nonDefaultFilters = pickBy(
+        filters,
+        (v, k: keyof FilterSelection) =>
+          !isEqual(v, DEFAULT_FILTER_SELECTION[k])
+      );
```
dajinchu: We used to encode the parameters into a URL string so we could pass it directly in the URL, but that's not needed now.

I would generate the object on line 93-96 inside `getKey` and `JSON.stringify` it. Then, in this fetcher function, just `JSON.parse` it and pass it to the `gqlClient`. Then we don't this odd intermediate representation of the url query string.

components/ResultsPage/useSearch.ts
```diff
@@ -99,3 +116,32 @@ export default function useSearch({
...
 }
+
+function transformGraphQLToSearchResult(
+  graphqlResults: SearchResultsQuery
+): SearchResult {
+  const transformedResults: SearchResult = {
+    results: [],
+    filterOptions: graphqlResults.search.filterOptions as FilterOptions,
+  };
+  transformedResults.results = graphqlResults.search.nodes.map((node) => {
```
dajinchu, line 115: This is a great approach! Adapting from graphql type to existing types lets us continue using the rest of the components without changes. 

As a future TODO, it might make sense to convert the rest of the code to use the original graphql data representation, and we can totally delete the `SearchResult` type, and just use the types from the generated graphql types. This would be less types/complexity and also potentially more performant since we can avoid this data transformation here. However, I'm not sure how much time it would really save us.

dajinchu (approved):
awesome job! Just fix the SWR function and we're good to go.

PR #86, opened 2021-03-05
Make GraphQL Endpoint Environment Variable

megandouli:
# what
GraphQL endpoint was hardcoded in and used in two places. We want to make it an environment variable and change it to localhost:4000 when doing full stack dev work.

# how 
Made the endpoint url an environment variable and replace where it was used. To load the environment variable in for the graphql-codegen, specified the `dotenv_config_path` in the command `generate:graphql`. To change the environment variable for full stack dev work, added another command that overrides the default graphql endpoint with `localhost:4000`.


utils/courseAPIClient.ts
```diff
@@ -1,7 +1,7 @@
 import { GraphQLClient } from 'graphql-request';
 import { getSdk } from '../generated/graphql';
 
-const ENDPOINT = 'https://searchneu.com/graphql';
```
dajinchu, line 4: This gets used in the frontend, so we need to tell Next.js to include this environment variable in the public bundle (because server-side stuff should be secret, but client side stuff is unavoidably public). 
https://nextjs.org/docs/basic-features/environment-variables

So just rename the env var to `NEXT_PUBLIC_GRAPHQL_ENDPOINT`.

dajinchu (changes requested):
Just rename the env var and we're good!

utils/courseAPIClient.ts
megandouli: done!

dajinchu (approved):
perfect

PR #88, opened 2021-03-06
Preload Pages

dajinchu:
This makes the term selector dropdown faster in production. It will still be slow in dev.
Using `next/link` components to preload the different home page variants. Link tags in the viewport are automatically preloaded. We could also preload using `router.preload` but I think it's neater to use `Link` and `a` tags instead. It also lets the browser to show the link location in the bottom left when you hover.


NEUDitao (approved):
nice

PR #92, opened 2021-03-14
Class Pages Part 1

megandouli:
# what
Started class pages!

# how
- [x] structured where class pages live so they can be accessed at a URL that looks like `/{campus}/{termId}/classPage/{subject}/{classId}/{searchQuery}`
- [x] Since both search result pages and class pages have a header that contains the logo, search bar, and campus & termId dropdown, refactored the header into its own component
- [x] Fix a bug with the mobileSearchOverlay where clicking the close icon or "View All Results" doesn't make the overlay disappear
- [x] Change the behavior of `setTermAndCampus` so it replaces parts of the url path instead of always going to a searchResult page, thus making the Header reusable by classPages

# prereqs
This builds off of #83 and should be merged after that PR.


dajinchu:
In `/{campus}/{termId}/classPage/{subject}/{classId}/{searchQuery}` 

what is the `searchQuery`? Shouldn't `NEU/202130/classPage/CS/2500` be enough to figure out which class to render?

megandouli:
> In `/{campus}/{termId}/classPage/{subject}/{classId}/{searchQuery}`
> 
> what is the `searchQuery`? Shouldn't `NEU/202130/classPage/CS/2500` be enough to figure out which class to render?

@dajinchu we have a "Back To Search Results" button on the class page that should return to the search results page the user was on previously with the searchQuery they had

dajinchu:
@megandouli It would be ideal not to put it in the path, so that we can cache the class pages (excellent for performance). The back to search could be done using `router.back` https://nextjs.org/docs/api-reference/next/router#routerback I believe?

components/Header.tsx
```diff
@@ -0,0 +1,166 @@
+import _ from 'lodash';
```
NEUDitao: big ol nit: when using `lodash`, can you use import the parts that you're using separately? theoretically could one day reduce bundle size

components/ResultsPage/Results/useResultDetail.tsx
```diff
@@ -73,7 +73,8 @@ export default function useResultDetail(
     const isCompositeReq = (
       variableToCheck: any
     ): variableToCheck is CompositeReq =>
-      (variableToCheck as CompositeReq).type !== undefined;
+      (variableToCheck as CompositeReq).type === 'and' ||
```
NEUDitao: isn't their a type for the `and` and `or`? ... might've been wiped with the old `Class` objects

megandouli:
@dajinchu oh cool we can definitely use `router.back` thanks for the tip! If we want the user's previous query to show up in the search bar for class pages, can we achieve that without putting `searchQuery` in the path?

components/Header.tsx
```diff
@@ -0,0 +1,166 @@
...
+    (t: string, newCampus: string) => {
+      const paths = window.location.pathname.split('/');
+      paths[1] = newCampus;
+      paths[2] = t;
+      router.push(paths.join('/'));
+    },
+    [router, query]
+  );
+
+  if (!termId || !campus) return null;
```
NEUDitao, line 65: is this needed? in what case are we rendering a `Header` where there's not a `termId` or `campus`? I don't mind it -- sincethis allows us to just place `Header` on any page and have the page handle whether or not to render it but just curious

pages/[campus]/[termId]/classPage/[subject]/[classId]/[query].tsx
```diff
@@ -0,0 +1,18 @@
...
+import React from 'react';
+import Header from '../../../../../../components/Header';
+
+export default function Page() {
+  const router = useRouter();
+
+  const termId = router.query.termId as string;
+  const campus = router.query.campus as string;
+  const subject = router.query.subject as string;
+  const classId = router.query.classId as string;
```
Comment (NEUDitao) [inline pages/[campus]/[termId]/classPage/[subject]/[classId]/[query].tsx]: food for thought for the future, we should make this `courseId` not `classId`, to be more consistent (and keep gradually wiping away the use of `class`)

NEUDitao (approved):
looks good, just some nits, and Dajin's comment

components/ResultsPage/Results/useResultDetail.tsx
megandouli: The type for `CompositeReq` didn't have a type for its type property
```
export interface CompositeReq {
  type: 'and' | 'or';
  values: Requisite[];
}
```

components/Header.tsx
megandouli: true we're doing a lot of the same checks in the components that use `Header` so I guess we can remove those

components/Header.tsx
megandouli: yup good point

dajinchu:
> @dajinchu oh cool we can definitely use `router.back` thanks for the tip! If we want the user's previous query to show up in the search bar for class pages, can we achieve that without putting `searchQuery` in the path?

yeah, put it in the query params like `?q=searchquery`

megandouli:
> > @dajinchu oh cool we can definitely use `router.back` thanks for the tip! If we want the user's previous query to show up in the search bar for class pages, can we achieve that without putting `searchQuery` in the path?
> 
> yeah, put it in the query params like `?q=searchquery`

@dajinchu got it I've moved classPages so they're on the path `/{campus}/{termId}/classPage/{subject}/{classId}` and if we call the query param `query` the `Header` component will work just like before

components/Header.tsx
```diff
@@ -0,0 +1,157 @@
...
+      `/${campus}/${termId}/search/${encodeURIComponent(q)}${
+        window.location.search
+      }`
+    );
+  };
+
+  const termAndCampusToURL = useCallback(
+    (t: string, newCampus: string) => {
+      const paths = window.location.pathname.split('/');
+      paths[1] = newCampus;
```
dajinchu: the substitution using the index will work but it feels rather brittle. I think it would make sense to accept `termAndCampusToURL` as a prop to Header, so that we can just do a simple string interpolation in the 2 places Header is used.

dajinchu (approved):

PR #93, opened 2021-03-20
Fix Frontend Rendering of Min/Max Credits

megandouli:
# what 
When min credits and max credits are different for a class, the frontend would render something like `1-1 Credits` which is wrong

# fix
Render `minCredits - maxCredits` if values are different


NEUDitao (approved):
😳

PR #98, opened 2021-03-24
graphql clear all bug fix

NEUDitao:
fixed the graphQL clear all bug fix by defaulting class range with an empty object


dajinchu:
great

dajinchu (approved):

PR #101, opened 2021-03-25
mutate user locally

dajinchu:


NEUDitao (approved):

utils/useUser.ts
```diff
@@ -43,19 +45,23 @@ export default function useUser(): UseUserReturn {
...
     const courseHash = Keys.getClassHash(course);
-    if (user?.followedCourses?.includes(courseHash)) {
+    if (followedCourses.includes(courseHash)) {
       macros.error('user already watching class?', courseHash, user);
       return;
     }
 
+    mutate(
+      { followedSections, followedCourses: [...followedCourses, courseHash] },
+      false
```
dajinchu, line 55: setting false tells SWR to just mutate the data locally without refetching

utils/useUser.ts
```diff
@@ -43,19 +45,23 @@ export default function useUser(): UseUserReturn {
...
+    );
     await subscribeToCourseUsingHash(courseHash);
     mutate();
```
dajinchu, line 58: this mutate tells swr to actually go refetch. So basically we set the new data locally, send the post request, and then refetch from server to make sure we get the actual correct data

PR #103, opened 2021-03-27
Class Pages General Info Section

megandouli:
# what 
[See Figma Design](https://www.figma.com/file/Eh1nMa7vQJ0mxSNIHvyI51/%E2%9C%85-Dev-Ready?node-id=807%3A1758)
The first section of the Figma design for class pages is done (general info). Class pages also show coreq class page info.

# how
- [x] Make a PageContent component (reused by the actual class page info as well as coreqs class page info)
- [x] Refactor out some common logic between SearchResultPages and ClassPages (like last updated string, credits display)


andrewydai (approved):
LGTM, only thing is a few lint errors and we've discussed this but down the road it'd be nice to take some of the logic heavy functions for PageContent and maybe put them somewhere else so that the component file is all mostly tsx code.

PR #105, opened 2021-04-07
Class page requisites section

puneetjohal:
This PR adds the requisites section of class pages. It also separates out sections into their own components from the `PageContent` file, and separates React-less functions into a service file.


components/ClassPage/ClassPageReqsBody.tsx
```diff
@@ -0,0 +1,250 @@
...
+              </div>
+            )
+          }
+        />
+      </div>
+    </div>
+  );
+}
+
+function PrereqsDisplay({
```
andrewydai: Still wondering whether this should be brought out as it's own component, maybe not even in the class-pages folder. Even if it's only used here, this feels class-page agnostic and is a lot of logic and code.

components/ResultsPage/Results/useResultDetail.tsx
```diff
@@ -3,23 +3,48 @@ import React, { ReactElement } from 'react';
...
+  coreqs: CompositeReq;
+}
+
+export type OptionalDisplay = (
+  PreqreqType: PrereqType,
+  Course: CourseReqs
+) => ReactElement | ReactElement[];
+
+export const isCompositeReq = (
+  variableToCheck: any
```
andrewydai: Maybe we could use unknown here instead of any? Unsure as I'm not a TS expert but I think any is strongly not preferred in TS practice.

components/ResultsPage/Results/useResultDetail.tsx
```diff
@@ -3,23 +3,48 @@ import React, { ReactElement } from 'react';
...
+): variableToCheck is CompositeReq =>
+  (variableToCheck as CompositeReq).type === 'and' ||
+  (variableToCheck as CompositeReq).type === 'or';
+
+export const isCourseReq = (
+  variableToCheck: any
```
andrewydai: Same as above

components/ResultsPage/Results/useResultDetail.tsx
```diff
@@ -3,23 +3,48 @@ import React, { ReactElement } from 'react';
...
 ): {
-  optionalDisplay: (PreqreqType, Course) => ReactElement | ReactElement[];
+  optionalDisplay: OptionalDisplay;
 } {
   const router = useRouter();
-  const onReqClick = (reqType, childBranch, event, searchQuery): void => {
+  const onReqClick = (reqType, childBranch, event, subject, classId): void => {
     router.push(
-      `/${router.query.campus}/${router.query.termId}/search/${searchQuery}`
+      `/${router.query.campus}/${router.query.termId}/classPage/${subject}/${classId}`
```
andrewydai: Just making sure, do we want this to occur? As in direct to class-page from search page or wherever else would use ```useResultDetail```

andrewydai (changes requested):
There are a few linting errors on code we wrote that I think can be pretty easily fixed up. For the most part looks good!

andrewydai (approved):

PR #106, opened 2021-04-08
Fix infinite scroll bug when searching exact course

dillonhammer:
# what

Fix the bug where if you search an exact course, and your window is small enough to have a scrollbar, then scrolling down would keep appending that one course to your results infinitely.

# how

Extract the `hasNextPage` flag from search queries and pass that as the `hasMore` prop to the `<InfiniteScroll>` component instead of hardcoding it to true as it was previously.


NEUDitao (approved):
thank you

PR #108, opened 2021-04-13
Tooltips

amielmon:
Added tooltips back. Here are screenshots:

Screen Shot 2021-04-13 at 7 37 16 PM

Screen Shot 2021-04-13 at 7 40 54 PM

Screen Shot 2021-04-13 at 7 36 48 PM

Screen Shot 2021-04-13 at 7 36 16 PM

Screen Shot 2021-04-13 at 7 47 24 PM


megandouli (changes requested):
Can you fix the failing typescript checks, front-end tests, and merge from master so you have the most recent changes? Not sure why the Vercel deployment is failing

components/types.ts
```diff
@@ -62,7 +62,7 @@ export interface Course {
   termId: string;
   desc: string;
   name: string;
-  prettyUrl: string;
```
andrewydai, line 65: Just wondering what this change is for.

pages/404.tsx
```diff
@@ -5,7 +5,7 @@ import Four04 from '../components/icons/404.svg';
 export default function NotFoundPage(): ReactElement {
   return (
     <div className="four04-container">
-      <div className="four04-title-text">something's borked</div>
+      <div className="four04-title-text">something&apos;s borked</div>
```
andrewydai: Good catch on the 404 page. I was reading online about this but it seems like ```&#39;``` is preferred. 
[https://stackoverflow.com/questions/419718/html-code-for-an-apostrophe](https://stackoverflow.com/questions/419718/html-code-for-an-apostrophe)
[http://www.madore.org/~david/computers/unicode/htmlent.html](http://www.madore.org/~david/computers/unicode/htmlent.html)
Let me know what you think

megandouli (approved):
very cool 👍

components/types.ts
amielmon: Oops added this back. I mistakenly thought that this field was no longer being returned from the backend because the globe links weren't working before (I think they broke sometime during the rewrite) but this must have been due to a typo.

pages/404.tsx
amielmon: I'm used to using '&apos;' and similar since its more mnemonic but I'll use your suggestion since it has HTML 4 support.

PR #117, opened 2021-08-11
Ad refactor banner alerts

andrewydai:
# Purpose

This pr refactors the AlertBanner component to draw data from a yaml file in the public folder. This is to reduce the need to do inline code changes to add banners. Going forward, front page alert banners can be added by modifying the `alert-banners.yml` file.

<br>

# Tickets

- My bad, no ticket :( 

# Contributors

- Andrew Dai

# Feature List

- Adds the ability to load and use yaml files throughout the project
- Adds a yaml file for easier adding and changing of banners in the future
- alert-banners.yml supports modifying alert level, text, visibility, and links for alerts
- Can easily render multiple alerts now

# Notes

<br>

# Checklist

- [x] Is passing tsc

<br>
@sandboxnu/searchneu

### DIFF
--- a/components/common/AlertBanner.tsx
+++ b/components/common/AlertBanner.tsx
@@ -3,28 +3,33 @@ import IconClose from '../icons/IconClose';
 
 type AlertLevel = 'error' | 'warning' | 'info';
 
-type AlertBannerProps = {
+export type AlertBannerData = {
+  text: string;
   alertLevel: AlertLevel;
-  alertMessage: string;
+  link?: string;
 };
 
-export default function AlertBanner({
-  alertLevel,
-  alertMessage,
-}: AlertBannerProps) {
-  const [isBannerOpen, setIsBannerOpen] = useState(true);
+type AlertBannerProps = {
+  alertBannerData: AlertBannerData;
+};
 
-  const closeBanner = () => setIsBannerOpen(false);
+export default function AlertBanner({ alertBannerData }: AlertBannerProps) {
+  const [isVisible, setIsVisible] = useState(true);
 
   return (
-    isBannerOpen && (
-      <div className={`alertBanner ${alertLevel}Banner`}>
-        {alertMessage}
+    isVisible && (
+      <div className={`alertBanner ${alertBannerData.alertLevel}Banner`}>
+        <span>
+          {alertBannerData.text}
+          {alertBannerData.link && (
+            <a href={alertBannerData.link}> Learn More.</a>
+          )}
+        </span>
         <div
           className="alertBanner__back"
           role="button"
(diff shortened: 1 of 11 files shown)


components/common/AlertBanner.tsx
```diff
@@ -1,34 +1,52 @@
...
-        >
-          <IconClose fill="#808080" />
-        </div>
-      </div>
-    )
+    <div className="alertBannerContainer">
+      {alertBanners.map((alertBanner) => {
+        return (
+          alertBanner.isVisible && (
+            <div className={`alertBanner ${alertBanner.alertLevel}Banner`}>
```
megandouli: make sure you add a `key` prop to the `div` since you're mapping

package-lock.json
```diff
@@ -0,0 +1,17194 @@
+{
+  "name": "searchneu-nextjs",
```
megandouli: we shouldn't have both `package-lock.json` and `yarn-lock`. It looks like you installed dependencies using `npm install` which is what generated the `package-lock.json`. You should delete this file and reinstall the dependnecies using `yarn install`

components/common/AlertBanner.tsx
```diff
@@ -1,34 +1,52 @@
...
 
-  const closeBanner = () => setIsBannerOpen(false);
+export default function AlertBanner() {
+  // this use state is used to trigger a rerender whenever closeBanner is called, without setting an entire
+  // object on state. Instead, we directly mutate alertBanners and call for a rerender by changing value.
+  const [value, setValue] = useState(0);
+
+  const closeBanner = (alertBanner: AlertBanner) => {
+    alertBanner.isVisible = false;
+    setValue(value + 1);
```
megandouli: I'm kinda confused about the design here. To me, it'd make more sense if the `AlertBanner` component renders 1 banner, so instead of mapping over all the banner data here, do it in `pages/[campus]/[termId].tsx` where you're actually rendering all the banners. 

I also don't understand why `value` is in state besides forcing a re-render when the user closes the banner. In that case, it seems like `isVisible` is what should actually be on state (default it to `props.isVisible`).

megandouli (changes requested):

components/common/AlertBanner.tsx
andrewydai: good catch, done

package-lock.json
andrewydai: changed

components/common/AlertBanner.tsx
andrewydai: changed

megandouli (approved):
awesome!

PR #120, opened 2021-08-27
Twilio

dillonhammer:
# Purpose

Rework notifications to use Twilio

<br>

# Tickets

https://trello.com/c/CxbutQVM/209-move-our-notification-system-away-from-facebook-notifications-to-twilio

# Contributors

@dillonhammer 
@megandouli 

# Feature List

- Add a Phone # verification modal when a user clicks to sign up for notifications
- Refactor existing notification toggles to send requests to new backend endpoint
- Remove all Facebook-related code

# Notes

Overall notification flow in `docs/Notifications.md`
New environment vars:
`NEXT_PUBLIC_NOTIFS_ENDPOINT`

<br>

# Checklist

- [ ] Approved by designers
- [x] Is passing tsc

<br>
@sandboxnu/searchneu


package.json
```diff
@@ -3,29 +3,25 @@
...
   },
   "dependencies": {
     "@babel/plugin-proposal-decorators": "^7.12.12",
-    "@prisma/client": "2.13.0",
     "@types/axios": "^0.14.0",
     "@types/jsonwebtoken": "^8.5.0",
     "@types/pg": "^7.14.7",
     "@wojtekmaj/enzyme-adapter-react-17": "^0.4.1",
     "amplitude-js": "^7.3.3",
+    "antd": "^4.16.9",
```
andrewydai, line 24: Can we use semantic-ui for out of the box stuff like buttons? It's already in the project and I think we should avoid adding external libraries if they're redundant. Alternatively, is there a specific reason we chose to use antd components for this pr? This comment is for all the spots that use antd

components/Header.tsx
```diff
@@ -23,19 +23,31 @@ import {
...
 type HeaderProps = {
   router: NextRouter;
   title: string;
   searchData: SearchResult;
   termAndCampusToURL: (t: string, newCampus: string, query: string) => string;
+  userInfo: UserInfo;
+  onSignOut: () => void;
 };
 
+export interface UserInfo {
```
andrewydai: can we move this to `types.ts` for consistency? I feel it was a common issue in the past for types to live in odd files and would be difficult to find. If there's no reason, I'd like to move it to `types.ts`

components/PhoneModal.tsx
```diff
@@ -0,0 +1,160 @@
...
+import { BarLoader } from 'react-spinners';
+
+interface PhoneModalProps {
+  visible: boolean;
+  onCancel: () => void;
+  onSignIn: (token: string) => void;
+  onSuccess: () => void;
+}
+
+export function PhoneModal({
```
andrewydai, line 16: I notice a lot of functions, useStates, and general bulkiness to this component. I think it would be beneficial to  either make it a class component (and with state and methods on the class), or make it have a service file of some sort to increase readability

components/PhoneModal.tsx
```diff
@@ -0,0 +1,160 @@
...
+      .post(`${process.env.NEXT_PUBLIC_NOTIFS_ENDPOINT}/sms/signup`, {
+        phoneNumber: `+${countryCode}${phoneNumber}`,
+      })
+      .then(() => {
+        setPhoneValidationMessage('');
+        if (!submitted) setSubmitted(true);
+        setResendDisabled(true);
+        setTimeout(() => setResendDisabled(false), 30 * 1000);
+      })
+      .catch((error) => {
```
andrewydai, line 43: Also think we should fix all the linting warnings in this file

components/PhoneModal.tsx
```diff
@@ -0,0 +1,160 @@
...
+        setLoading(false);
+        setResendDisabled(false);
+        setResponseMessage(
+          'Error: Please try again or request a new verification code.'
+        );
+      });
+  };
+
+  const onCountryCodeChange = (value: any): void => {
```
andrewydai: Could maybe consolidate some of these validator functions into one number validator function that takes an argument, a function (state setter), and maybe a regex (but that doesn't look necessary right now).

components/PhoneModal.tsx
andrewydai: For the event.target.value typing which is being used as `any` in this file, can maybe use something like `HTMLInputElement`. Unsure if this works though, have to test a little

components/PhoneModal.tsx
```diff
@@ -0,0 +1,160 @@
...
+  const [submitted, setSubmitted] = useState(false);
+  const [resendDisabled, setResendDisabled] = useState(false);
+  const [verificationCode, setVerificationCode] = useState('');
+  const [loading, setLoading] = useState(false);
+  const [phoneValidationMessage, setPhoneValidationMessage] = useState('');
+  const [responseMessage, setResponseMessage] = useState('');
+
+  const onPhoneNumberSubmit = (): void => {
+    axios
+      .post(`${process.env.NEXT_PUBLIC_NOTIFS_ENDPOINT}/sms/signup`, {
```
andrewydai, line 34: Might be nice to have some additional regex validation here so we don't need to send a request to the backend to know we have an issue. Thoughts?

components/PhoneModal.tsx
```diff
@@ -0,0 +1,160 @@
...
+        setResendDisabled(false);
+        setResponseMessage(
+          'Error: Please try again or request a new verification code.'
+        );
+      });
+  };
+
+  const onCountryCodeChange = (value: any): void => {
+    const reg = /^\d*$/;
+    if (!isNaN(value) && reg.test(value)) {
```
andrewydai: I think this is redundant. Aren't these regex's only verifying each character is a numeric value (which isNaN does too)? I think it'd be good to have a regex that allows for `()` and `-` since those are common

components/ResultsPage/Results/NotifSignUpButton.tsx
```diff
@@ -0,0 +1,12 @@
+import React, { ReactElement } from 'react';
+import { Button } from 'antd';
+
+interface NotifSignUpButtonProps {
+  onNotifSignUp: () => void;
+}
+
+export default function NotifSignUpButton({
```
andrewydai, line 8: I think this button is maybe too simple to be a component. It basically provides users a button that they can only control the `onClick` prop for and has unchangeable text. I think it's probably fine to just make this one line inline element in the `SignUpForNotifications.tsx`

pages/[campus]/[termId]/search/[query].tsx
```diff
@@ -105,6 +140,9 @@ export default function Results(): ReactElement | null {
               results={searchData.results}
               loadMore={loadMore}
               hasNextPage={searchData.hasNextPage}
+              userInfo={userInfo}
```
Comment (andrewydai) [inline pages/[campus]/[termId]/search/[query].tsx:145]: I am a little concerned about the passing of all these props down the massive component tree. The 2 reasons I am worried are:
1. This bloats code, as every intermediary component between this component and the SignUpForNotifications component has to add these to their props list and their props interfaces
2. This could cause problems in performance as changes on this data can cause a ripple of rerenders on our state tree when we really only want to rerender parts of it.
In terms of solutions, I would reference this [React context](https://reactjs.org/docs/context.html#when-to-use-context). React context are a bit involved and maybe too heavy handed here, composition might be the way to go. Essentially, in composition we would pass down the SignUpForNotifications component from the top to where it's needed (inversion of control). 

It might not be the right thing to do here, LMK what you think I'm open to suggestions.

docs/Notifications.md
```diff
@@ -1,33 +1,52 @@
...
+  - TWILIO_PHONE_NUMBER
+  - TWILIO_VERIFY_SERVICE_ID
+  - CLIENT_ORIGIN
+  - JWT_SECRET
+- Frontend:
+  - NEXT_PUBLIC_NOTIFS_ENDPOINT
+
+## Sending Notifications
+
+TODO: @Megan
```
andrewydai, line 41: Is this still a TODO or does this refer to the prod setup of our twilio endpoints? I thought the backend code was done for texting

andrewydai (changes requested):

megandouli (commented):
leaving a comment so I don't forget, we should remove the FB alert and add a new one about SMS notifications

docs/Notifications.md
megandouli: yeah the backend code isn't totally done, just have the minimal stuff out. I'll create a ticket for these things.

components/PhoneModal.tsx
megandouli: hmm there is a good amount of state on this component but I didn't find it very hard to read or understand, and splitting it up might just make it harder to figure out how pieces are connected so I'm leaning towards not breaking this component up.

components/PhoneModal.tsx
megandouli: I think we can use `string` for `target.value` instead of `any`.

components/PhoneModal.tsx
megandouli: country codes should only ever by numbers so I think this regex check is fine... however, it's true that valid phone numbers _could_ have `()` and `-` so I think we're going to remove the regex check in `onPhoneChange` and basically allow them to put in whatever, then perform a regex check before submitting the phone number to the backend, stripping out the extra valid characters.

components/PhoneModal.tsx
megandouli: yee

components/ResultsPage/Results/NotifSignUpButton.tsx
megandouli: true, it is a very simple component right now but I can imagine it getting a little more complicated when we implement Diana's designs, so I'd prefer to leave it as it is.

kiminkim724:
I changed onPhoneNumberSubmit so that the code automatically adds the dashes for the phone number as they input each number. So the users are only allowed to input numbers, which means that the only cases that would return an invalid phone number would be when the phone number isn't long enough or it's an invalid phone number. I was wondering if I should create a separate function to check for the phone number length or just have an if statement at the top of onPhoneNumberSubmit to check for this?

components/PhoneModal.tsx
```diff
@@ -0,0 +1,189 @@
...
+
+  const onCountryCodeChange = (value: string): void => {
+    const reg = /^\d*$/;
+    if (!isNaN(parseInt(value)) && reg.test(value)) {
+      setCountryCode(value);
+    }
+  };
+
+  const onPhoneChange = (value: string): void => {
+    console.log(`Phone number: ${value}`);
```
andrewydai: Remove

components/PhoneModal.tsx
```diff
@@ -0,0 +1,189 @@
...
+    const reg = /^[\d-]*$/;
```
andrewydai: Maybe we should move all these regex's outside of the component to a constant at the top of this file. Some are reused, but also redeclaring them every function call seems akin to doing `new RegExp()` everytime as per this [post](https://stackoverflow.com/questions/9750338/dynamic-vs-inline-regexp-performance-in-javascript/32523333) which could be expensive

components/PhoneModal.tsx
```diff
@@ -0,0 +1,189 @@
...
+    const val = value.split('-').join('');
+    if (
+      value == '' ||
```
andrewydai: Always use triple equals with JS/TS. Double equals is strange, and will return true for things like `1 == "1"`. I'm honestly surprised we don't have a linter to yell at this lol

components/PhoneModal.tsx
```diff
@@ -0,0 +1,189 @@
...
+      (!isNaN(parseInt(value)) &&
```
andrewydai: NaN is falsy, meaning in a boolean expression it is false. In this case, parseInt() will give NaN if it is not a number, so you don't need isNaN.

andrewydai: Actually, this may give 0 itself which is falsy, so maybe keep this.

components/PhoneModal.tsx
```diff
@@ -0,0 +1,189 @@
...
+        }
+        setPhoneNumber(value);
+      }
+      console.log(value.length);
+      value.length == 12 ? setResendDisabled(false) : setResendDisabled(true);
+    }
+  };
+
+  // checks if the given Phone Number has dashes in the correct places
+  function checkPhoneNumberDashes(value: string): boolean {
```
andrewydai: Let's try to keep style in the same, especially because this is already in a function body, aka `const functionName = () =>`. Also, I prefer predicate functions be named as a function, `isPhoneNumberDashesValid` for example.

components/PhoneModal.tsx
```diff
@@ -0,0 +1,189 @@
...
+    console.log(`Phone number: ${value}`);
+    const reg = /^[\d-]*$/;
+    const val = value.split('-').join('');
+    if (
+      value == '' ||
+      (!isNaN(parseInt(value)) &&
+        reg.test(value) &&
+        checkPhoneNumberDashes(value))
+    ) {
+      if (value == '') {
```
andrewydai: Isn't this redundant?

components/PhoneModal.tsx
```diff
@@ -0,0 +1,189 @@
...
+        setPhoneNumber(value);
+      } else {
```
andrewydai: Can replace with `else if() {` on same line

components/PhoneModal.tsx
andrewydai: Also, unsure if this is necessary given the invariant restriction above and that users can enter dashes. At least from what i understand, let me know

components/PhoneModal.tsx
```diff
@@ -0,0 +1,189 @@
...
+        if (val.length < 9) {
+          value = val.match(/.{1,3}/g).join('-');
+        }
+        setPhoneNumber(value);
+      }
+      console.log(value.length);
```
andrewydai: remove

components/PhoneModal.tsx
```diff
@@ -0,0 +1,189 @@
...
+      value.length == 12 ? setResendDisabled(false) : setResendDisabled(true);
```
andrewydai: In general with this function, i'm a little lost what the logic flow is. Maybe some comments or walk through would help me.

andrewydai (changes requested):
Overall, the changes implemented are good, but the logic around the phone is iffy to me, and there's some syntax/convention issues that should be cleaned up. LMK if you have questions!

components/PhoneModal.tsx
kiminkim724: I think the regex check should already check if it's a number, so honestly, I think the whole isNaN check might be redundant? Any thoughts on that?

components/PhoneModal.tsx
kiminkim724: I got rid of the function, I found a better way to implement the dashes in the phone number.

components/PhoneModal.tsx
kiminkim724: I realized the problem was when I was calling the regexp to add dashes in the phone number wouldn't work when the length is zero, so I just checked if the length was greater than zero in the statement below.

components/PhoneModal.tsx
kiminkim724: I've cleaned up the code so it's a lot simpler, I'll try to leave some comments to explain what some of the parts do.

components/PhoneModal.tsx
```diff
@@ -0,0 +1,175 @@
...
+}: PhoneModalProps): ReactElement {
+  const [countryCode, setCountryCode] = useState('1');
+  const [phoneNumber, setPhoneNumber] = useState('');
+  const [submitted, setSubmitted] = useState(false);
+  const [resendDisabled, setResendDisabled] = useState(true);
+  const [verificationCode, setVerificationCode] = useState('');
+  const [loading, setLoading] = useState(false);
+  const [phoneValidationMessage, setPhoneValidationMessage] = useState('');
+  const [responseMessage, setResponseMessage] = useState('');
+  const phoneReg = /^\d*$/;
```
andrewydai: sorry this is nitpicky, let's do this outside the PhoneModal function but not with the imports, just before this function. And use all caps to signify a constant.

andrewydai (approved):

PR #125, opened 2021-10-07
Add total aggregation value in filter headers

puneetjohal:
# Purpose

Add total aggregation value in filter headers to improve user understanding of how many results are being displayed. The aggregation is equal to the number of results being displayed because of the selected filters. The total aggregation value is omitted when there are no filters selected. 

<br>

# Tickets

https://trello.com/c/65DClxKh

# Contributors

- @puneetjohal 

# Notes

Expected behavior with no filters applied:
image

Expected behavior with filters applied:
image

Expected behavior when the user searches for something else without clearing filters and the selected filters no longer apply to any search results (ex. you search for "japanese buddism", apply the JPNS and CLTR subject filters, and then search for "cs2500" without clearing filters):
image

<br>

# Checklist

- [x] Approved by designers
- [x] Is passing tsc
- [ ] ~~Has documentation and comments in code~~
- [ ] ~~Has test coverage for code~~

<br>
@sandboxnu/searchneu


megandouli (changes requested):
I don't think this behavior makes sense... I select the `CS` subject filter which has 69 results, so I'd expect the aggregation value to be 69 instead of 1478. @beccajohnson17 what do you think?

Screen Shot 2021-10-09 at 5 04 18 PM

beccajohnson17:
I agree with @megandouli that once a filter is chosen, then the header should update to show the running total of all filters selected. So it would say 69, but if EECE was added, it would be 114. Also a side question but wouldn't the Class Type also have a running total?

puneetjohal:
Waiting on designer input for this one.

utils/filterUtils.ts
```diff
@@ -0,0 +1,12 @@
+export function getTotalAggregation(selected, options): string {
```
megandouli, line 1: nit: I think it makes more sense for this function to return a number instead of a formatted string. The name implies it returns a total so it shouldn't also be concerned with string formatting

utils/filterUtils.ts
```diff
@@ -0,0 +1,12 @@
...
+  if (selected.length > 0) {
+    let selectedOptions = options.filter((option) =>
```
megandouli, line 3: just make this a const to remove the lint error

megandouli (approved):
looks good! just 2 super minor things

puneetjohal:
We're ditching this design, I'll put in a new PR for the new design.

PR #134, opened 2021-11-07
Remove hardcoded term IDs

hankewyczz:
# Purpose

###### Remove hardcoded term IDs - get them programmatically from the backend

<br>

# Tickets

###### https://trello.com/c/xRH6xepp/216-remove-hardcoded-termids

-

# Contributors

###### Zachar Hankewycz, Kimin Kim

-

# Feature List

###### Adds support for asyncronously fetching the term IDs from the backend (and updating the list asyncronously across all components

-
-
-

# Notes

###### N/A

<br>

# Checklist

- [ ] Approved by designers
- [ ] Is passing tsc
- [ x] Has documentation and comments in code

<br>
@sandboxnu/searchneu

### DIFF
--- a/components/Header.tsx
+++ b/components/Header.tsx
@@ -2,7 +2,7 @@ import { merge } from 'lodash';
 import Head from 'next/head';
 import Link from 'next/link';
 import { NextRouter } from 'next/router';
-import React, { ReactElement, useCallback } from 'react';
+import React, { ReactElement, useCallback, useEffect, useState } from 'react';
 import { BooleanParam, useQueryParam, useQueryParams } from 'use-query-params';
 import { getRoundedTerm, getTermInfoForCampus } from '../components/global';
 import FilterButton from '../components/icons/FilterButton.svg';
@@ -49,6 +49,34 @@ export default function Header({
   const query = (router.query.query as string) || '';
   const termId = router.query.termId as string;
   const campus = router.query.campus as string;
+
+  // Get the lists of terms asyncronously
+  const [termInfos, setTermInfos] = useState([]);
+  // Get the rounded terms asyncronously
+  const [searchOptions, setSearchOptions] = useState([]);
+
+  useEffect(() => {
+    // Update the lists of terms
+    getTermInfoForCampus(Campus[campus.toUpperCase()])
+      .then((data) =>
+        data.map((terminfo) => ({
+          text: terminfo.text,
+          value: terminfo.value,
+          link: termAndCampusToURLCallback(terminfo.value, campus),
+        }))
+      )
+      .then((data) => setTermInfos(data));
+
+    let campusOptions = Object.keys(Campus).map((c: Campus) =>
+      getRoundedTerm(c, termId).then((term) => ({
+        text: c,
+        value: c,
+        link: termAndCampusToURLCallback(term, c),
+      }))
+    );
+
+    Promise.all(campusOptions).then((data) => setSearchOptions(data));
+  }, []);
 
   const [qParams, setQParams] = useQueryParams(QUERY_PARAM_ENCODERS);
   const filters: FilterSelection = merge({}, DEFAULT_FILTER_SELECTION, qParams);
@@ -127,11 +155,7 @@ export default function Header({
         <div className="Breadcrumb_Container">
           <div className="Breadcrumb_Container__dropDownContainer">
             <SearchDropdown
-              options={Object.keys(Campus).map((c: Campus) => ({
-                text: c,
-                value: c,
-                link: termAndCampusToURLCallback(getRoundedTerm(c, termId), c),
-              }))}
+              options={searchOptions}
               value={campus}
               className="searchDropdown"
               compact={false}
@@ -140,13 +164,7 @@ export default function Header({
           <span className="Breadcrumb_Container__slash">/</span>
           <div className="Breadcrumb_Container__dropDownContainer">
             <SearchDropdown
-              options={getTermInfoForCampus(Campus[campus.toUpperCase()]).map(
-                (terminfo) => ({
-                  text: terminfo.text,
-                  value: terminfo.value,
-                  link: termAndCampusToURLCallback(terminfo.value, campus),
-                })
-              )}
+              options={termInfos}
               value={termId}
               className="searchDropdown"
               compact={false}
--- a/components/Header.tsx
+++ b/components/Header.tsx
@@ -2,7 +2,7 @@ import { merge } from 'lodash';
 import Head from 'next/head';
 import Link from 'next/link';
 import { NextRouter } from 'next/router';
-import React, { ReactElement, useCallback, useContext } from 'react';
+import React, { ReactElement, useCallback } from 'react';
 import { BooleanParam, useQueryParam, useQueryParams } from 'use-query-params';
 import { getRoundedTerm } from './terms';
 import FilterButton from '../components/icons/FilterButton.svg';
@@ -25,7 +25,7 @@ import {
 import { campusToColor } from '../utils/campusToColor';
 import MobileSearchOverlay from './ResultsPage/MobileSearchOverlay';
 import { Button } from 'antd';
-import { termsContext } from '../utils/TermInfoProvider';
+import getTermInfos from '../utils/TermInfoProvider';
 
 type HeaderProps = {
   router: NextRouter;
@@ -52,7 +52,7 @@ export default function Header({
   const campus = router.query.campus as string;
 
   // Get the TermInfo dict from the app context
-  const termInfos = useContext(termsContext);
+  const termInfos = getTermInfos();
 
   const [qParams, setQParams] = useQueryParams(QUERY_PARAM_ENCODERS);
   const filters: FilterSelection = merge({}, DEFAULT_FILTER_SELECTION, qParams);
--- a/components/Header.tsx
+++ b/components/Header.tsx
@@ -51,13 +51,7 @@ export default function Header({
   const termId = router.query.termId as string;
   const campus = router.query.campus as string;
 
-  const termAndCampusToURLCallback = useCallback(
-    (t: string, newCampus: string) => {
-      return termAndCampusToURL(t, newCampus, query);
-    },
-    [query, termAndCampusToURL]
-  );
-
+  // Get the TermInfo dict from the app context
   const termInfos = useContext(termsContext);
 
   const [qParams, setQParams] = useQueryParams(QUERY_PARAM_ENCODERS);
@@ -70,6 +64,13 @@ export default function Header({
       }`
     );
   };
+
+  const termAndCampusToURLCallback = useCallback(
+    (t: string, newCampus: string) => {
+      return termAndCampusToURL(t, newCampus, query);
+    },
+    [query, termAndCampusToURL]
+  );
 
   if (!termId || !campus) return null;
   if (showOverlay && macros.isMobile) {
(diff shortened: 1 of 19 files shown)


components/Header.tsx
```diff
@@ -50,6 +50,41 @@ export default function Header({
...
+
+  // The list of termIDs (for the selected campus) is updated every time the campus var changes
+  useEffect(() => {
+    // Update the lists of terms
+    getTermInfoForCampus(Campus[campus.toUpperCase()])
+      .then((data) =>
+        data.map((terminfo) => ({
+          text: terminfo.text,
+          value: terminfo.value,
+          link: termAndCampusToURLCallback(terminfo.value, campus),
```
andrewydai: To fix the linting issue pointing here, you need to move the `termAndCampusToURLCallback` definition above these two useEffect hooks

components/HomePage/ExploratorySearchButton.tsx
```diff
@@ -1,7 +1,8 @@
-import React, { ReactElement } from 'react';
+import React, { ReactElement, useState, useEffect } from 'react';
 import { Campus } from '../types';
 import { useRouter } from 'next/router';
 import { getTermName } from '../global';
+import SectionsTermNav from '../ClassPage/SectionsTermNav';
```
andrewydai: remove

components/HomePage/HomeSearch.tsx
```diff
@@ -16,11 +16,50 @@ interface HomeSearchProps {
...
+    const campusLinksDict = {};
+    // Get the data we need, map it, and insert it into the dictionary
+    // This pattern isn't intuitive, but Promise.all can't resolve on a dictionary object, so we have to work around that
+    const campusLinksTemp = Object.values(Campus).map((c) =>
+      getRoundedTerm(c, termId)
+        .then((term) => `/${c}/${term}`)
+        .then((url) => (campusLinksDict[c] = url))
+    );
+
+    Promise.all(campusLinksTemp).then((data) =>
```
andrewydai: remove arg from lambda callback

components/ResultsPage/SearchDropdown.tsx
```diff
@@ -1,5 +1,5 @@
 import Link from 'next/link';
-import React, { ReactElement } from 'react';
+import React, { ReactElement, useState } from 'react';
```
andrewydai: remove useState

andrewydai (changes requested):
General comment: I recognize that the whole system of components depending on hardcoded termids was difficult to change but something doesn't fully sit well with the useEffect useState pattern I'm seeing everywhere. Wonder if a smarter way to do this is on the whole app container (_app) we wait for it to mount, and then immediately request for every termid we can. When that finishes, we set a global readable state on react context that everyone can update off of when it changes, and show the correct values. Meanwhile everywhere else displays empty values until that comes in. 

Looking for other devs thoughts too on this

components/Header.tsx
```diff
@@ -50,6 +50,48 @@ export default function Header({
...
+    (t: string, newCampus: string) => {
+      return termAndCampusToURL(t, newCampus, query);
+    },
+    [query, termAndCampusToURL]
+  );
+
+  // Get/set some termID-related variables asyncronously
+  const [termInfos, setTermInfos] = useState([]);
+  const [roundedTerms, setRoundedTerms] = useState([]);
+
```
megandouli: would prefer to move these `useState`s towards the top of the component so someone looking at the code can see what state the component has immediately

components/HomePage/HomeSearch.tsx
```diff
@@ -16,11 +16,48 @@ interface HomeSearchProps {
 }
 
 const HomeSearch = ({ termId, campus }: HomeSearchProps): ReactElement => {
+  // Get the lists of terms asyncronously
+  const [termInfos, setTermInfos] = useState([]);
+
+  // Ditto, for the Campus links
+  const emptyLinks = {};
+  Object.values(Campus).map((c) => (emptyLinks[c] = ''));
```
megandouli: better style to create an obj with keys from an array would be to use `.reduce` instead of `.map` so something like 

```
const emptyLinks = Object.values(Campus).reduce((acc, curr) => {...body function}, {});
```

components/global.ts
```diff
@@ -1,84 +1,87 @@
...
-    default:
-      return [];
+      subCollege = 'LAW';
+      break;
   }
+
+  // Grab the data from GQL
+  const data = await gqlClient
+    .getTermIDsByCollege({ subCollege: subCollege })
+    .then((res) => res['termInfos']);
```
megandouli: I don't think you need to do `await` and `.then()` here, you should be able to just do 

```
const data = (await gqlClient
    .getTermIDsByCollege({ subCollege: subCollege })).termInfos;
``

File: components/global.ts [hunk 8]
```diff
@@ -1,84 +1,87 @@
...
-    case Campus.CPS:
-      return cpsTerms[0].value as string;
-    case Campus.LAW:
-      return lawTerms[0].value as string;
-    default:
-      return '';
+/**
+Queries the TermInfos for this campus from the backend, or returns a cached version if we've already done so.
+
+// TODO - Decide how often we should force-rescrape (Otherwise, we won't re-hit the backend until the next time the program is resarted.)
```
megandouli: I wouldn't call this `force-rescrape`... it'd be confusing since the backend has scrapers. Really, this is about refetching data from the backend. Why not just let each frontend client make the request once per campus? Instead of keeping a cache in memory, you could memoize `getTermInfoFromBackend` using `memoize` from `lodash` and use the campus as the cache key. Also maybe rename `getTermInfoFromBackend` to `fetchTermInfo`

components/HomePage/ExploratorySearchButton.tsx
```diff
@@ -13,15 +13,26 @@ const ExploratorySearchButton = ({
...
 }: ExploratorySearchButtonProps): ReactElement => {
   const router = useRouter();
+
+  const [termName, setTermName] = useState('');
+
+  // The term name should be updated every time either the campus or the term ID changes
+  useEffect(() => {
+    getTermName(termId).then((t) => {
+      // Check if there's a matching name
+      if (typeof t !== 'undefined') {
```
megandouli: why are we checking the typeof t isn't undefined instead of the value of t? Also when would `t` be undefined?

components/HomePage/HomeSearch.tsx
```diff
@@ -16,11 +16,48 @@ interface HomeSearchProps {
...
+  // Update the list of links to each campus every time the termID changes
+  // Each link depends on the rounded term (the term closest to the main termID, but for a given campus)
+  useEffect(() => {
+    const campusLinksDict = {};
+    // Get the data we need, map it, and insert it into the dictionary
+    // This pattern isn't intuitive, but Promise.all can't resolve on a dictionary object, so we have to work around that
+    const campusLinksTemp = Object.values(Campus).map((c) =>
+      getRoundedTerm(c, termId)
+        .then((term) => `/${c}/${term}`)
+        .then((url) => (campusLinksDict[c] = url))
```
megandouli: this doesn't need to be 2 `.then()`s; you can convert the `term` to `url` in one function body

components/HomePage/HomeSearch.tsx
```diff
@@ -16,11 +16,48 @@ interface HomeSearchProps {
...
+  useEffect(() => {
+    getTermInfoForCampus(campus)
+      .then((data) =>
+        data.map((terminfo) => ({
+          text: terminfo.text,
+          value: terminfo.value,
+          link: `/${campus}/${terminfo.value}`,
+        }))
+      )
+      .then((data) => setTermInfos(data));
```
megandouli: don't need 2 `.then()`s, conversion from `data` to list of `termInfos` can happen in one `.then`

components/Header.tsx
```diff
@@ -50,6 +50,48 @@ export default function Header({
...
+    // Update the lists of terms
+    getTermInfoForCampus(Campus[campus.toUpperCase()])
+      .then((data) =>
+        data.map((terminfo) => ({
+          text: terminfo.text,
+          value: terminfo.value,
+          link: termAndCampusToURLCallback(terminfo.value, campus),
+        }))
+      )
+      .then((data) => setTermInfos(data));
```
megandouli: pretty sure you can do all this inside one `then` see below for similar comments

components/HomePage/HomeSearch.tsx
```diff
@@ -16,11 +16,48 @@ interface HomeSearchProps {
...
+  }, [campus]);
+
+  // Update the list of links to each campus every time the termID changes
+  // Each link depends on the rounded term (the term closest to the main termID, but for a given campus)
+  useEffect(() => {
+    const campusLinksDict = {};
+    // Get the data we need, map it, and insert it into the dictionary
+    // This pattern isn't intuitive, but Promise.all can't resolve on a dictionary object, so we have to work around that
```
megandouli: there's gotta be a better way to do this but I'm too tired to think of it

components/global.ts
```diff
@@ -1,84 +1,87 @@
...
+// This dictionary maps Campus -> (array of termIDs)
+const termInfoCache: Record<string, TermInfo[]> = {};
+for (const c of Object.values(Campus)) {
+  termInfoCache[c] = [];
+}
+
+/**
+ * Queries the backend for information about the terms for the given campus
+ */
+async function getTermInfoFromBackend(c: Campus): Promise<TermInfo[]> {
```
megandouli: I'd probably rename this file to `terms.ts` at this point

pages/api/sitemap.xml.ts
```diff
@@ -30,10 +30,10 @@ async function generateSitemap(): Promise<string> {
   const items: Set<string> = new Set();
 
   // latest terms for each campus
-  const latestTerms: [Campus, string][] = Object.values(Campus).map((c) => [
-    c,
-    getLatestTerm(c),
-  ]);
+  let latestTerms: [Campus, string][];
```
megandouli: you can use `p-map` (there are more examples of its usage in course-catalog-api) if you're mapping over something and the mapper function is async. Alternatively since the `.map` returns a Promise, you can `Promise.all` a `.map`

megandouli (changes requested):
overall agree with Andrew; all the `useStates` and `useEffects` don't seem appropriate since really we just want to make 1 request to the backend to fetch termIds. I'm not a React expert, so I can't say what the right way to do this is... would encourage you to ask in #engineering-support in Slack

components/terms.ts
```diff
@@ -0,0 +1,169 @@
...
+  text: string;
+  /** Term ID */
+  value: string;
+  href: string;
+}
+
+/**
+ * Queries the backend for information about the terms for all campuses
+ */
+export async function fetchTermInfo(): Promise<{ [key: string]: TermInfo[] }> {
```
megandouli: minor Typescript thing, instead of `{[key: string]: TermInfo[]}` to represent an object where the key is a string and the value is a list of TermInfos, use `Record<string, TermInfo[]>`. If you can be more specific with your key, that'd be even better. In this case, I think it'd be `Record<Campus, TermInfo[]>`.

Note that `Record<Campus, TermInfo[]>` means this type needs to have every key in `Campus`. If that's not the case, you can use `Partial<Record<Campus, TermInfo[]>>` which means the type does not need every key in `Campus` but if it does have a key, the key is a `Campus`.

pages/[campus]/[termId].tsx
```diff
@@ -97,7 +93,10 @@ export default function Home(): ReactElement {
           >
             <div className="centerTextContainer">
               <Logo className="logo" aria-label="logo" campus={campus} />
-              <HomeSearch termId={termId} campus={campus} />
+              {termInfos[campus].length == 0 && <LoadingContainer />}
+              {termInfos[campus].length > 0 && (
+                <HomeSearch termId={termId} campus={campus} />
+              )}
```
Comment (megandouli) [inline pages/[campus]/[termId].tsx:100]: unless `termInfos[campus]` could be undefined, these two blocks can be combined into a ternary so something like

```
{ termInfos[campus].length === 0 ? <LoadingContainer />  : <HomeSearch termId={termId} campus={campus} /> }
```

utils/TermInfoProvider.tsx
```diff
@@ -0,0 +1,28 @@
+import React, { useState, useEffect } from 'react';
+import { fetchTermInfo, TermInfo } from '../components/terms';
+import { Campus } from '../components/types';
+import PropTypes from 'prop-types';
+
+const emptyTermInfos: { [key: string]: TermInfo[] } = Object.values(
```
megandouli: use the type from my other comment here

megandouli (approved):
just small things to fix; fantastic job!!!

andrewydai (changes requested):
Great work! There are a few linting warnings that should be addressed, and I also think it could be nice to export a function on the TermInfoProvider that does `return useContext(termsContext)` so you don't have to import useContext and termsContext everywhere.

NEUDitao (approved):

PR #145, opened 2021-12-03
Fix 'undefined undefined' for course page prereqs

hankewyczz:
# Purpose

<br>

 On class pages, text-based (as opposed to courses) prereqs show up as undefined undefined. This is not an issue with search results.

# Tickets

-https://trello.com/c/4O0h9Cih/280-graduation-admission-dissertation-check-show-up-as-undefined-undefined-on-class-pages

# Contributors

-Zachar

# Notes

<br>

Before:
image

After:
image

# Checklist

- [ ] Approved by designers
- [ ] Is passing tsc

<br>
@sandboxnu/searchneu

### DIFF
--- a/components/ClassPage/PrereqsDisplay.tsx
+++ b/components/ClassPage/PrereqsDisplay.tsx
@@ -1,7 +1,10 @@
 import Link from 'next/link';
 import React, { ReactElement } from 'react';
-import { isCompositeReq } from '../ResultsPage/Results/useResultDetail';
-import { CompositeReq, CourseReq, Requisite } from '../types';
+import {
+  isCompositeReq,
+  isCourseReq,
+} from '../ResultsPage/Results/useResultDetail';
+import { CompositeReq, Requisite } from '../types';
 
 type PrereqsDisplayProps = {
   termId: string;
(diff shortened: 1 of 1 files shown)


kiminkim724 (approved):
Looks good!

components/ClassPage/PrereqsDisplay.tsx
```diff
@@ -70,13 +73,18 @@ export default function PrereqsDisplay({
...
-      <li key={prereq.subject + prereq.classId}>
+      <li key={prereqs.subject + prereqs.classId}>
         <Link
-          href={`/${campus}/${termId}/classPage/${prereq.subject}/${prereq.classId}`}
-        >{`${prereq.subject} ${prereq.classId}`}</Link>
+          href={`/${campus}/${termId}/classPage/${prereqs.subject}/${prereqs.classId}`}
+        >{`${prereqs.subject} ${prereqs.classId}`}</Link>
+      </li>
+    );
+  } else {
```
kiminkim724, line 84: Actually on second thought, you check if prereqs is a string in an else if statement in useResultDetail.tsx. Do you want to do the same thing here? 
else if (typeof prereqs === 'string') instead of just else?

kiminkim724 (changes requested):

components/ClassPage/PrereqsDisplay.tsx
megandouli: @kiminkim724 that shouldn't be necessary if we're not doing type casting (which is what line 74 used to be). TypeScript is smart enough to infer types based off control flow so the type of `prereq` is a `Requisite` which is a union of `CompositeReq`, `CourseReq`, and `string`. If we reach the else case, TypeScript will know it's a string so it should be safe. If the type of `Requisite` ever changes, TypeScript will probably complain

megandouli (approved):
:yay:

PR #158, opened 2022-03-10
add integration test for home page

andrewydai:
# Purpose

This PR is meant to demonstrate what a new test would look like for searchneu. It cleans up our testing environment a little and adds an integration test for our front page, testing that user actions do invoke the correct things down the line. It will be continued with another integration test for results page.
<br>

# Tickets

# Contributors

- @andrewydai 

# Feature List

- Cleans up the act issue with state changes in tests
- some small package changes (script name change and added axios mock adapter)
- added an integration test for the home page and results page

# Notes

- the generated graphql file changed after I ran some update scripts to deal with a persistent warning from NPX. Not entirely sure why it changed though but my frontend still functions
<br>

# Checklist

- [ ] Approved by designers
- [x] Is passing tsc

<br>
@sandboxnu/searchneu


yarn.lock
```diff
@@ -5094,6 +5104,32 @@ es-abstract@^1.17.2, es-abstract@^1.17.4, es-abstract@^1.18.0, es-abstract@^1.18
     string.prototype.trimstart "^1.0.4"
     unbox-primitive "^1.0.1"
 
+es-abstract@^1.19.0, es-abstract@^1.19.1:
```
hankewyczz, line 5112: [nit] Could you install `yarn-deduplicate` and run it? 

We have it in CCA - just add as dev dependency.
https://github.com/sandboxnu/course-catalog-api/blob/master/package.json#L178

components/tests/integration/Home.test.jsx
```diff
@@ -0,0 +1,110 @@
...
+  });
+
+  afterEach(() => {
+    jest.clearAllMocks();
+    mockAxios.reset();
+  });
+
+  it('should push a query to the router when executing a searching', () => {
+    const homePage = mount(<Home />);
+    const searchBarInput = homePage.find('input');
```
hankewyczz, line 55: I think this needs a team conversation to pick a testing framework and stick with it, but I'm a fan of [`react-testing-library`](https://github.com/testing-library/react-testing-library).

It allows you to write tests like a user would navigate the page, instead of relying on element IDs and classes.

For example:
```
fireEvent.click(screen.getByLabelText(/View all classes/i))
```

andrewydai: OK we can quickly talk about this at standup and I can amend this PR depending on what we decide. I'll try writing some tests in `react-testing-library`

andrewydai: I ended up sticking with enzyme. The main reasons are its existence already in our codebase, and the lack of a good way to select elements without text. I think that while enzyme requires a more intimate knowledge of the rendered DOM once tests are setup they still provide the same value, and ideally in the future you write tests while developing you should already know how to work with the part you're testing.

hankewyczz (approved):

PR #161, opened 2022-04-06
switch total results to use total count

andrewydai:
# Purpose

Switches from calculating total results from aggregations in filters to totalCount field on the searchResults graphql API, which is based on total hits from ES
<br>

# Tickets

- Could not find one tho there may be one 🤔 

# Contributors

- @andrewydai 

# Feature List

- Switches total results to use total count
- Removes that function
- Formats the string 

# Notes

ES generally has a 10,000 result limit regardless of pagination, so if 10,000 results return I added a + marker.

<br>

# Checklist

- [ ] Approved by designers
- [x] Is passing tsc

<br>
@sandboxnu/searchneu

image

### DIFF
--- a/pages/[campus]/[termId]/search/[query].tsx
+++ b/pages/[campus]/[termId]/search/[query].tsx
@@ -100,40 +100,22 @@ export default function Results(): ReactElement | null {
     }`;
   };
 
-  const getTotalResults = (): number => {
-    /* 
-      Using the subject filter here to get the total number of search results because:
-      - the results themselves are loaded 10 at a time
-      - classes don't overlap across subjects
-    */
-    const options = searchData?.filterOptions['subject'] || [];
-    const selected = filters['subject'];
-
-    if (selected.length > 0) {
-      const selectedOptions = options.filter((option) =>
-        selected.includes(option.value)
-      );
-      return selectedOptions.reduce(
-        (total_aggregation, option) => total_aggregation + option.count,
-        0
-      );
-    }
-
-    return options.reduce(
-      (total_aggregation, option) => total_aggregation + option.count,
-      0
-    );
-  };
-
   const TotalResultsDisplay = (): ReactElement => {
-    const totalResults = getTotalResults();
-    return (
-      <p>
-        {totalResults === 1
-          ? `${totalResults} course result`
-          : `${totalResults} course results`}
-      </p>
-    );
+    // This has to be a null safe because searchData can be undefined on mount
+    const totalResults = searchData?.totalCount || 0;
+    let totalResultsStr = '';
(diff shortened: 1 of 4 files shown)


hankewyczz (approved):
🚢

pages/[campus]/[termId]/search/[query].tsx
```diff
@@ -100,40 +100,22 @@ export default function Results(): ReactElement | null {
...
-      </p>
-    );
+    // This has to be a null safe because searchData can be undefined on mount
+    const totalResults = searchData?.totalCount || 0;
+    let totalResultsStr = '';
+    switch (totalResults) {
+      case 1:
+        totalResultsStr = ' result';
+        break;
+      case 10000:
```
Comment (puneetjohal) [inline pages/[campus]/[termId]/search/[query].tsx:117]: why specifically 10,000 or was this an arbitrary choice

puneetjohal (approved):

pages/[campus]/[termId]/search/[query].tsx
Comment (hankewyczz) [reply pages/[campus]/[termId]/search/[query].tsx:117]: I kinda glossed over this until your comment, but what does this do? I'm very tired so I might just be missing something obvious, but won't this result in the following behavior:

```
9999 -> 9999 results
10000 -> 10000+ results
10001 -> 10001 results
```

Is that what you meant to do?

pages/[campus]/[termId]/search/[query].tsx
```diff
@@ -100,40 +100,22 @@ export default function Results(): ReactElement | null {
...
-    const totalResults = getTotalResults();
-    return (
-      <p>
-        {totalResults === 1
-          ? `${totalResults} course result`
-          : `${totalResults} course results`}
-      </p>
-    );
+    // This has to be a null safe because searchData can be undefined on mount
+    const totalResults = searchData?.totalCount || 0;
```
Comment (hankewyczz) [inline pages/[campus]/[termId]/search/[query].tsx]: I was on the fence about this earlier (and so didn't leave a comment), but now that I've thought about it - would it make more sense to just not render anything when `totalCount` is null? Otherwise, we conflate "zero results" and "the search hasn't returned yet", and I'm not a huge fan of that.

pages/[campus]/[termId]/search/[query].tsx
Comment (andrewydai) [reply pages/[campus]/[termId]/search/[query].tsx:117]: yea i can leave a comment about this but it's also described in pr description, TLDR elasticsearch has a 10,000 hit limit by default. We haven't changed it and I don't see a time where we do. The consequence of this is that the max value total count can be (which is based off of total hits) iss 10,000

Comment (hankewyczz) [reply pages/[campus]/[termId]/search/[query].tsx:117]: lol i read that earlier and it completely left my mind when I came back for the second look

PR #170, opened 2022-09-24
added waitlists seats to mobile

DilanRocco:
# Purpose
Added ability to view waitlist seats on mobile view. Previously, the waitlist seats would not be visible through the results page on mobile.
<br>

# Tickets

 - https://trello.com/c/K22NCJWp

# Contributors

- Dilan Piscatello

# Feature List

Screen Shot 2022-09-28 at 2 58 28 PM

-

# Notes

<br>

# Checklist

- [ ] Approved by designers
- [x] Is passing tsc

<br>
@sandboxnu/searchneu

### DIFF
--- a/styles/results/_MobileSectionPanel.scss
+++ b/styles/results/_MobileSectionPanel.scss
@@ -58,6 +58,7 @@
   }
   &__waitlist-text {
     color: gray;
+    white-space: no-wrap;
   }
 
   & > .green {
--- a/styles/results/_MobileSectionPanel.scss
+++ b/styles/results/_MobileSectionPanel.scss
@@ -56,10 +56,7 @@
       margin-bottom: 16px;
     }
   }
-  &__waitlist {
-    display: flex;
-    justify-content: space-between;
-    margin-top: -16px;
+  &__waitlist-text {
     color: gray;
   }
 
--- a/styles/results/_MobileSectionPanel.scss
+++ b/styles/results/_MobileSectionPanel.scss
@@ -56,6 +56,12 @@
       margin-bottom: 16px;
     }
   }
+  &__waitlist {
+    display: flex;
+    justify-content: space-between;
+    margin-top: -16px;
+    color: gray;
+  }
 
   & > .green {
     margin-top: 16px;
--- a/styles/results/_MobileSectionPanel.scss
+++ b/styles/results/_MobileSectionPanel.scss
@@ -57,8 +57,8 @@
     }
   }
   &__waitlist-text {
-    color: gray;
-    white-space: no-wrap;
+    color: #969696;
+    white-space: nowrap;
   }
 
   & > .green {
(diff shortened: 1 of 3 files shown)


styles/results/_MobileSectionPanel.scss
```diff
@@ -56,6 +56,10 @@
       margin-bottom: 16px;
     }
   }
+  &__waitlist-text {
```
hankewyczz, line 59: this class doesn't seem to have any effect - the `span` elements don't actually inherit these styles.

hankewyczz (changes requested):

styles/results/_MobileSectionPanel.scss
DilanRocco: the class __waitlist-text has two css attributes: color, and text-wrap. At least visually, I can see that the waitlist seats changes color to gray. Am I missing something else?

styles/results/_MobileSectionPanel.scss
```diff
@@ -56,6 +56,10 @@
...
+    color: gray;
```
koulopoulos: nit: instead of using `gray` here, we probably want to use an existing shade like `#969696`(the color we use for 'None' for NUPaths/Prerequisites/Corequisites/Course fees)

styles/results/_MobileSectionPanel.scss
```diff
@@ -56,6 +56,10 @@
...
+    white-space: no-wrap;
```
koulopoulos: Does `no-wrap` work here with the hyphen? ime the value needs to be `nowrap` (see [https://developer.mozilla.org/en-US/docs/Web/CSS/white-space](https://developer.mozilla.org/en-US/docs/Web/CSS/white-space))

koulopoulos (changes requested):
Just a couple of styling nitpicks 👍

styles/results/_MobileSectionPanel.scss
hankewyczz: ah, you're right about the color - I was just looking at the wrapping behavior. In this case, following Adam's comment should be fine.

DilanRocco: understood

styles/results/_MobileSectionPanel.scss
DilanRocco: understood, right now I have the color hard coded, not sure if we want to put that into a constants file

styles/results/_MobileSectionPanel.scss
DilanRocco: yes, you're right. Fixed.

hankewyczz (approved):
lgtm

styles/results/_MobileSectionPanel.scss
koulopoulos: should be fine hard coded for now

koulopoulos (approved):
🚀🚀🚀

PR #182, opened 2022-11-06
Negative Seat Count Tooltip

Lucas-Dunker:
# Purpose

Adds a tooltip for _negative_ seatcounts explaining why they are negative (a course advisor set an override for a student to take a course)
<br>

# Tickets

[-Trello Link](https://trello.com/c/F9Qz43pi/313-info-tooltip-for-negative-seat-count)

# Contributors

@Lucas-Dunker 
@sebwittr 

# Feature List

image

-Adds tooltip for negative seat counts explaining why they are negative
-Changed tooltip color for all tooltips to better fit SearchNEU's color palette
-Added better padding for all tooltips

# Notes

*

<br>

# Checklist

- [x] Approved by designers
- [x] Is passing tsc

<br>

Primary Reviewer: @hankewyczz 
Secondary Reviewers: @pranavphadke1 @allychao 

@sandboxnu/searchneu


components/common/NegativeSeatCountTooltip.tsx
```diff
@@ -0,0 +1,14 @@
+import React, { ReactElement } from 'react';
+import IconInfo from '../icons/info-icon.svg';
+import Tooltip, { TooltipDirection } from '../Tooltip';
+
+const TOOLTIP_MESSAGE =
+  'Negative seat counts are displayed when students override into a course.';
+export default function SearchInfoIcon() {
```
hankewyczz: Shouldn't this be `NegativeSeatCount...` or similar? Not `SearchInfo..`

components/common/NegativeSeatCountTooltip.tsx
```diff
@@ -0,0 +1,14 @@
...
+  return (
+    <div className="SearchInfoIcon">
```
hankewyczz: naming - since this isn't related to the search info, can we find a more general name for that icon?

styles/_SearchInfoIcon.scss
```diff
@@ -3,12 +3,13 @@
   &__icon {
     &:hover + .tooltip {
       display: block;
-      background: #338ef1;
+      background: #1d3557;
```
hankewyczz: Whenever making styling changes like this, make sure to talk to Vanessa to get her signoff on it (and make a note of it in the PR)

styles/results/_DesktopSectionPanel.scss
```diff
@@ -49,25 +49,33 @@
     }
   }
 
-  & > td:nth-child(3) {
+  > td:nth-child(3) {
```
hankewyczz, line 38: what does this change do?

styles/results/_DesktopSectionPanel.scss
```diff
@@ -49,25 +49,33 @@
...
+    > div {
+      display: flex;
+      flex-direction: row;
+      justify-content: space-between;
+      align-items: center;
+      width: 25%;
     }
 
+    line-height: 20px;
+
```
hankewyczz: what do these changes do?

components/ResultsPage/Results/SectionPanel.tsx
```diff
@@ -171,9 +172,12 @@ export function DesktopSectionPanel({
       </td>
       <td>{section.campus}</td>
       <td>
-        <span className={getSeatsClass()}>
-          {section.seatsRemaining}/{section.seatsCapacity}
-        </span>
+        <div className="DeskopSectionPanel__seatcount">
```
hankewyczz, line 178: Can y'all talk to Vanessa - should we show something similar on mobile search results? We can't do a tooltip, but maybe just a text element whenever a section has a negative seat count?

hankewyczz (changes requested):
Looks good for the most part, just some questions/minor changes. 

A general note - when making styling changes that aren't universal (in this case - only applicable to specific sections), it's always helpful to include a link in the PR. 

For example - here's a search link for `POLS4500`, Fall 2022 semester, which has negative seats left: https://searchneu-git-ldnegativetooltip-sandboxneu.vercel.app/NEU/202310/search/POLS4500

Also - that link above cuts off the tooltip for me. Could it maybe be aligned to the left, so it doesn't get cut off?

image

styles/results/_DesktopSectionPanel.scss
sebwittr: I think the & is just unnecessary for that one

components/ResultsPage/Results/SectionPanel.tsx
sebwittr: We got her approval for all the other changes, we'll ask about what we should do on mobile.

styles/results/_DesktopSectionPanel.scss
sebwittr: CSS was giving us weird problems with having these classes where they previously were, since we changed the HTML and wrapped them inside a div (Basically it wasn't respecting the styling anymore, since there was another layer). I don't think the red, yellow, or green classes were being used elsewhere so I thought the best solution was to bring them outside the styling of the td, which fixed our problem.

styles/_SearchInfoIcon.scss
sebwittr: Vanessa approved this color change!

components/ResultsPage/Results/SectionPanel.tsx
Lucas-Dunker: > Looks good for the most part, just some questions/minor changes.
> 
> A general note - when making styling changes that aren't universal (in this case - only applicable to specific sections), it's always helpful to include a link in the PR.
> 
> For example - here's a search link for `POLS4500`, Fall 2022 semester, which has negative seats left: https://searchneu-git-ldnegativetooltip-sandboxneu.vercel.app/NEU/202310/search/POLS4500
> 
> Also - that link above cuts off the tooltip for me. Could it maybe be aligned to the left, so it doesn't get cut off?
> 
> image

The textbox is only cut off at smaller resolutions, which Seb and I believe is expected; we messed around with having the tooltip box expand with text to the left instead of the right, but we couldn't get it working unfortunately. We're worried that messing with the box sizing might mess up the arrow at the top of the box as well.

hankewyczz: "expected" as in "this is desired behavior" or "given how the code is now, this is how it works"?

Either way — we don't want to leave it in this state. Don't modify the existing `Tooltip`, but expand it. An idea could be to add a `prop` which allows us to optionally define which direction (left/right) the tooltip is oriented in.

Lucas-Dunker: Great suggestion, we'll expand `prop` and get some direction functionality working.

components/Tooltip.tsx
```diff
@@ -5,16 +5,17 @@ export enum TooltipDirection {
   Down = 'DOWN',
 }
 
-type TooltipProps = {
+export type TooltipProps = {
   text: string;
   direction: TooltipDirection;
+  flipLeft?: boolean;
```
hankewyczz, line 16: Could this be a little more descriptive? We want a name that explains that if this flag is `true`, then the body of the tooltip will be oriented to the left of the tooltip arrow. (edit — I see that this is a prop passed directly to the Tooltip, so maybe you can't change the name. If that's the case, you'll still want to document its use)

That's a little hard to fit into a variable name, so you'll definitely want to add a JSDoc comment right above that variable, in the props. (JSDoc > normal comment, because in most IDEs, if you hover over a variable/attribute/property/function with a JSDoc comment, you'll see it in a little tooltip. That's not the case w/ normal comments)

components/ResultsPage/Results/SectionPanel.tsx
```diff
@@ -280,8 +290,17 @@ export function MobileSectionPanel({
           <span> | </span>
           <span className="MobileSectionPanel__waitlist-text">
             {`${section.waitRemaining}/${section.waitCapacity} Waitlist Seats`}
-          </span>
+          </span>{' '}
```
hankewyczz: Why do we need this?

EDIT: if we really need this space, I'd prefer if you just added a non-breaking space (`&nbsp;`) in the inner span (ie. right before "Negative"), just to keep those two related components together.

hankewyczz (commented):
99.9%!!! I just have two tiny nits

components/ResultsPage/Results/SectionPanel.tsx
```diff
@@ -171,9 +172,18 @@ export function DesktopSectionPanel({
...
-        <span className={getSeatsClass()}>
-          {section.seatsRemaining}/{section.seatsCapacity}
-        </span>
+        <div className="DeskopSectionPanel__seatcount">
+          <span className={'DesktopSectionPanel__' + getSeatsClass()}>
+            {section.seatsRemaining}/{section.seatsCapacity}
+          </span>
+          {section.seatsRemaining < 0 && (
+            <InfoIconTooltip
+              text="Negative seat counts are displayed when students override into a course."
```
hankewyczz: I'd like to expand this text a little bit, to add more context (what is an "override"?)

So, two options:
- Add a link to the word "override" to [this article](https://registrar.northeastern.edu/article/student-registration-overrides/)
  - Right now, the tooltip component takes text, not HTML elements. However, this should be a relatively straightforward change
- Change the text phrasing to make it clear that an "override" is something that an academic advisor does on behalf of a student. For example: "Negative seat counts are displayed when academic advisors apply course overrides for students."

I'm fine with either one; the second option is definitely easier, so I'm fine with you guys doing that.

hankewyczz (changes requested):
https://searchneu-7qan0f8ps-sandboxneu.vercel.app/NEU/202330/search/is4300

image

The tooltip looks off-center for me - could we change that?

Lucas-Dunker:
New Alignment: image

New Color Scheme + Alignment:  
image

DilanRocco:
The padding between the two lines of text have an abnormal amount of space. At least on the current website, one is right under the other.
Screenshot 2022-12-04 at 4 47 43 PM

components/ResultsPage/Results/SectionPanel.tsx
```diff
@@ -178,7 +178,7 @@ export function DesktopSectionPanel({
           </span>
           {section.seatsRemaining < 0 && (
             <InfoIconTooltip
-              text="Negative seat counts are displayed when students override into a course."
```
hankewyczz: The mobile version should also be updated. I recommend making a top-level variable for the negative seat tooltip text, so that there is only one place to change the text.

hankewyczz (changes requested):
One more tiny thing

hankewyczz:
@Lucas-Dunker 

Small CSS tweak:
image

we've got a little clipping on the tooltip icon (but only at certain zoom levels — not sure why).

- e: can be fixed by increasing the width and height of the SVG (`InfoIconTooltip__icon`) to `16px`

Also, the tooltip isn't fully aligned with the text box

hankewyczz (approved):

sebwittr (approved):

PR #185, opened 2022-11-08
Semesters Names Inconsistency on Law/CPS Class Pages

tiingweii-shii:
# Purpose

<br> CPS/LAW use different semester names than NEU, but it is not reflected on course info pages. This PR addresses the semester name inconsistencies :)

The logic to get semester names was changed: we decide the semester name from its matching termId from termInfos of corresponding campus (college).
Screen Shot 2022-11-20 at 4 38 01 PM

# Tickets

https://trello.com/c/G5MKa3nX/260-semesters-for-law-cps-classes-dont-match-up-on-class-pages

# Contributors

@tiingweii-shii 

# Feature List

- Updated the logic for getting semester names from termIds (to display correct semester names for CPS and LAW)

# Notes
- For some classes, banner gives sections for semesters that are not included in termInfos, so we need to filter out those sections. However, it is best to handle this in the backend. A ticket will be cut for this. See comment for details 👀
- Currently the CourseInfo page breaks (with a `LocationUndefined` error) whenever it is refreshed. Cabbage says Seb is working on a ticket that is related to this. 

# Checklist

- [ ] Approved by designers
- [x] Is passing tsc

<br>
@sandboxnu/searchneu

### DIFF
--- a/components/ClassPage/PageContent.tsx
+++ b/components/ClassPage/PageContent.tsx
@@ -5,6 +5,7 @@ import ClassPageInfoBody from './ClassPageInfoBody';
 import ClassPageInfoHeader from './ClassPageInfoHeader';
 import ClassPageReqsBody from './ClassPageReqsBody';
 import ClassPageSections from './ClassPageSections';
+import getTermInfos from '../../utils/TermInfoProvider';
 
 type PageContentProps = {
   termId: string;
@@ -28,8 +29,13 @@ export default function PageContent({
   // TODO: hacky front-end solution because for some reason allOccurrences includes
   // termIds where there are no sections. This should probably be fixed on the backend.
   if (classPageInfo && classPageInfo.class) {
+    const termIds = getTermInfos()[campus].map((termInfo) => termInfo.value);
     classPageInfo.class.allOccurrences = classPageInfo.class.allOccurrences.filter(
-      (occurrence) => occurrence.sections.length > 0
+      (occurrence) =>
+        occurrence.sections.length > 0 &&
+        // filter out termIds that are not availible from termInfos from Banner,
+        // probably should include this in the backend as well
+        termIds.includes(occurrence.termId)
     );
   }
   return (
--- a/components/terms.ts
+++ b/components/terms.ts
@@ -1,6 +1,7 @@
 import _ from 'lodash';
 import { Campus } from './types';
 import { gqlClient } from '../utils/courseAPIClient';
+import getTermInfos from '../utils/TermInfoProvider';
 
 /** Information about a term */
 export interface TermInfo {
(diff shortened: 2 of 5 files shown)


components/terms.ts
```diff
@@ -126,26 +126,44 @@ export function getTermName(
...
-export function getYear(termId: string): number {
+export function getYearFromTermId(termId: string): number {
   const givenYear: number = parseInt(termId.substr(0, 4));
-  const season = getSeason(termId);
-  if (season === 'Fall' || season === 'Winter') {
+  const seasonDigit: number = parseInt(termId.charAt(termId.length - 2));
+  // Fall and Winter semesters occurs in the previous year from the given year from termId.
+  // ex: 202110 should be Fall 2020 not Fall 2021
+  // Fall semesters have season digit 1; Winter semesters have season digit 2.
+  if (seasonDigit < 3) {
```
tiingweii-shii: I think "correcting" the year for Fall and Winter semesters is pretty thoughtful and it is consistent with the Homepage semester, but it seems like the Winter semester (for CPS and LAW) is not set correctly on Homepage (e.g. termId 202325 is Winter 2023, but it should be Winter 2022 to be consistent).

I may push another commit to fix this, but thats a problem for tomorrow...

components/terms.ts
```diff
@@ -126,26 +126,44 @@ export function getTermName(
...
+    '40': 'Summer I',
+    '50': 'Summer Full',
+    '54': 'Summer Semester',
+    '55': 'Summer Quarter',
+    '60': 'Summer II',
   };
-  if (!seasonDigitMap[seasonDigit]) throw new Error('unexpected season digit');
-  else return seasonDigitMap[seasonDigit];
+
+  if (!semesterDigitMap[semesterDigit])
```
hankewyczz:
```suggestion
  if (semesterDigitMap[semesterDigit] === undefined)
```

components/terms.ts
hankewyczz: We want to stay consistent with what Banner gives us [[link]](https://nubanner.neu.edu/StudentRegistrationSsb/ssb/classSearch/getTerms?offset=1&max=200&searchTerm=)

So, Fall 2022-23 is "Fall 2022", but Winter 2022-23 is "Winter 2023"

This function should only modify the Fall names.

hankewyczz (changes requested):
We have the TermInfoContext - could we use that to directly get the names instead of using a mapping?

components/ClassPage/PageContent.tsx
```diff
@@ -28,8 +29,13 @@ export default function PageContent({
   // TODO: hacky front-end solution because for some reason allOccurrences includes
   // termIds where there are no sections. This should probably be fixed on the backend.
   if (classPageInfo && classPageInfo.class) {
+    const termIds = getTermInfos()[campus].map((termInfo) => termInfo.value);
     classPageInfo.class.allOccurrences = classPageInfo.class.allOccurrences.filter(
-      (occurrence) => occurrence.sections.length > 0
+      (occurrence) =>
+        occurrence.sections.length > 0 &&
+        // filter out termIds that are not availible from termInfos from Banner,
+        // probably should include this in the backend as well
```
tiingweii-shii, line 37: This should be handled in the backend. Cabbage says he will cut a ticket for this.

Issue #186, opened 2022-11-09
Change upper limit of Course ID Range slider from 6000 to 8000


IsaacKhor:
There are many 6xxx and 7xxx courses (for us grad students), so it would be nice if we could filter for them without manually manipulating the parameters in the url.

PR #187, opened 2022-11-09
Add 7000 and 8000 ticks to the course id range filter

IsaacKhor:
# Purpose

As per #186, add the 7000 and 8000 ticks to the course id range filter slider so grad students can search for courses too.

# Contributors

Just me

# Notes

I haven't actually tested if the design *looks* any good, I literally `grep`'ed through the codebase, found the appropriate list, and added two numbers. Somebody with the whole dev environment setup should take a look at it first.

<br>

# Checklist

- [ ] Approved by designers
- [x] Is passing tsc

<br>
@sandboxnu/searchneu

### DIFF
--- a/components/ResultsPage/RangeFilter.tsx
+++ b/components/ResultsPage/RangeFilter.tsx
@@ -16,7 +16,7 @@ export default function RangeFilter({
 }: RangeFilterProps): ReactElement {
   const [controlledInput, setControlledInput] = useState(selected);
 
-  const courseIDs = [1000, 2000, 3000, 4000, 5000, 6000];
+  const courseIDs = [1000, 2000, 3000, 4000, 5000, 6000, 7000, 8000];
   const marks = {};
   courseIDs.forEach((id) => {
     marks[id] = id.toString();


hankewyczz (commented):
We originally limited the scale of the course slider to 0 - 6000 because going up to 9000 doesn't look great visually. 

The logic for cutting out 6000-9000 was based off the assumption that grad students have a more specific set of classes to take & won't be perusing classes within a range as often as undergrads. 

We'll take a look at this, thanks!

IsaacKhor:
> We originally limited the scale of the course slider to 0 - 6000 because going up to 9000 doesn't look great visually.

Yeah that's what I was worried about, the space seems a little tight there. We don't need to go up to 9000, I think a 7000 and 8000 should be enough.

> The logic for cutting out 6000-9000 was based off the assumption that grad students have a more specific set of classes to take & won't be perusing classes within a range as often as undergrads.

You'd be surprised. For CS, the they're changing the requirements for PhD candidacy to at least 4 7xxx classes and at most 2 5-6xxx, so it's actually quite common for us to just filter for 7000 and above to take a look around.

Hope this helps and provides some context around why this is a desirable feature :)

hankewyczz:
> We don't need to go up to 9000, I think a 7000 and 8000 should be enough.

Well, class IDs go up all the way to 9996 - if we're expanding to contain 7000 and 8000, we'd probably want to expand to fit all class IDs.

Thanks for the feedback - we didn't realize this was a common issue. I'm not sure if you've noticed, but Vercel deployed your PR [[link here]](https://searchneu-5hu8esjmy-sandboxneu.vercel.app). You can use this for the time being while we figure out a longer-term solution (which may take a while; we're decently busy right now). That link uses the same backend API as the main website does, so you should be able to use it without any issues.

IsaacKhor:
> Well, class IDs go up all the way to 9996 - if we're expanding to contain 7000 and 8000, we'd probably want to expand to fit all class IDs.

On the contrary, 8xxx and 9xxx are all "fake" courses that only exist so the system sees that we're full-time students. They're stuff like "readings" (pre-candidacy version of "work with your advisor") and "dissertation term 1" (i.e. work on your dissertation). The full listing is [here](https://catalog.northeastern.edu/graduate/academic-policies-procedures/records-transcripts/). Course 9996 is literally "dissertation continuation" for all colleges, which I would wager nobody would want listed here anyways.

So there's really no reason to go beyond 8000, since everything above that aren't "real" courses anybody would actually search for.

PR #198, opened 2023-01-29
Sort Section Search Results by Subscription then Open Seat Count

soulwa:
# Purpose

Currently, sections are sorted by professor names. It might be more helpful for users to sort by subscription and open seat count.

<br>

# Tickets

- https://trello.com/c/6ULqeoqP

# Contributors

- @soulwa 
- @hankewyczz 

# Feature List

- replace `sortSections` with a sort by subscription status, then remaining section seats, then remaining waitlist seats in `CourseResult.tsx`, have it take `userInfo` as necessary context for sorting by subscription.

# Checklist

- [ ] Approved by designers
- [ ] Is passing tsc

<br>
@sandboxnu/searchneu

### DIFF
--- a/components/ResultsPage/Results/CourseResult.tsx
+++ b/components/ResultsPage/Results/CourseResult.tsx
@@ -26,9 +26,7 @@ interface CourseResultProps {
 }
 
 const sortSections = (sections: Section[], userInfo?: UserInfo): Section[] => {
-  // NOTE (sam 2023-01-29): unsure why sections needs to be cloned, because all we're doing is sorting, nothing desctructive.
-  // also unsure why sections are memoized after being sorted (in CourseResult)
-  // memoized based on current course, but course should never change over lifetime of a CourseResult
+  // TODO (sam 2023-03-09): remove this `cloneDeep` call once we can remove the `useMemo` from `CourseResult`.
   const sortedSections = cloneDeep(sections);
   const subscribedSectionIds = new Set(userInfo?.sectionIds ?? []);
   sortedSections.sort((a: Section, b: Section) => {
@@ -57,6 +55,7 @@ export function CourseResult({
   const router = useRouter();
   const termId = router.query.termId as string;
   const campus = router.query.campus as string;
+  // TODO (sam 2023-03-09): this is necessary because of `useShowAll`, which should likely not be coupled to courses.
   const sortedSections = useMemo(
     () => sortSections(course.sections, userInfo),
     [course]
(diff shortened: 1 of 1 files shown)


components/ResultsPage/Results/CourseResult.tsx
```diff
@@ -48,7 +57,10 @@ export function CourseResult({
   const router = useRouter();
   const termId = router.query.termId as string;
   const campus = router.query.campus as string;
-  const sortedSections = useMemo(() => sortSections(course.sections), [course]);
+  const sortedSections = useMemo(
+    () => sortSections(course.sections, userInfo),
+    [course]
```
hankewyczz, line 61: We should add `userInfo` as a dependency (and the same for the lower `useMemo` instance)

soulwa: iirc we didn't do this because this would make classes which you registered for immediately appear at the top of the list (it would resort the list right away, rather than on refresh) -- is this behavior we want for the UI?

hankewyczz: ahh, sorry, I forgot about that. You're right

components/ResultsPage/Results/CourseResult.tsx
```diff
@@ -17,24 +17,33 @@ import { DesktopSectionPanel, MobileSectionPanel } from './SectionPanel';
...
   onSignIn: (token: string) => void;
   fetchUserInfo: () => void;
 }
 
-const sortSections = (sections: Section[]): Section[] => {
+const sortSections = (sections: Section[], userInfo?: UserInfo): Section[] => {
+  // NOTE (sam 2023-01-29): unsure why sections needs to be cloned, because all we're doing is sorting, nothing desctructive.
+  // also unsure why sections are memoized after being sorted (in CourseResult)
+  // memoized based on current course, but course should never change over lifetime of a CourseResult
   const sortedSections = cloneDeep(sections);
```
Lucas-Dunker: If we're unsure, would removing the clone be a separate PR or could we just include it in this one? Same for the memorization

soulwa: alright so- 
`cloneDeep` is necessary here because of `useMemo`, since `useMemo` require that its function argument be pure. *technically* we could get away with a shallow copy here, since this sort isn't (and never should be?) modifying the actual `Section` objects in the array. but for safety I'm not opposed to leaving the `cloneDeep`, as long as we leave the `useMemo`. I'm open to leaving it in at the moment in case we want to have course results reorder themselves on notification, since that is not a behavior we want now, but could maybe do something with in the future. 

so options are
1. leave `useMemo` and `cloneDeep` (guarantees pure)
2. leave `useMemo`, move to shallow copy (have to uphold invariant that sort doesn't modify objects, which it **shouldn't**, but also removes a `lodash` dependency)

Lucas-Dunker (approved):

soulwa:
finally had a chance to actually look into this `useMemo` business- I don't think it's necessary, but right now without it `useShowAll` causes hundreds of re-renders. I've got an idea for a way to decouple this, but to get this PR through I'm going to update the comment and leave it be.

pranavphadke1 (approved):

PR #199, opened 2023-01-29
Api error page

pranavphadke1:
# Purpose

###### Previously, whenever our API didn't work we would display a simple error message. Now we've switched to a new design that reroutes to an error page whenever the API isn't working.

On desktop:
image

On mobile:

https://user-images.githubusercontent.com/91645231/217370222-368ddd05-184a-4a51-964e-7892008a1e28.mov

<br>

# Tickets

-https://trello.com/c/vslDRINJ/348-implement-api-error-page-for-frontend

# Contributors

- @pranavphadke1 

# Feature List

- Uploaded new crying husky image (crying-husky-2.svg since crying-husky.svg already exists)
- Added a new error page. Redirect to it whenever the API fails. This is done through useEffect that checks if "termInfosError" is no longer null --> router push to new page
- Added scss page for the new error page. Imported it into base.scss
- Removed feedback modal from termId page since it is now on the apiError page
- Updated snapshot test (since main page no longer has the feedback modal)

# Notes

<br>

- I had trouble figuring out styling for when the page is accessed on a mobile device. However, I think it should be decent now.
- Also wasn't sure what to change for when the API fails during a search, so I left that how it was originally.
re: 
image

# Checklist

- [ ] Approved by designers
- [ ] Is passing tsc

<br>
@sandboxnu/searchneu


pranavphadke1:
Note: 
- Will probably use this error page as default. Should prob change the page name to something cleaner than "/apiError"
- I believe the only "error" we can get from the useSearch function in "pages/campus/termid/search/query.tsx" is "TypeError: Network request failed", which I believe can only happen if the API isn't working. If this is the case, maybe we should just redirect to this new error page in a similar fasion as we do in the homepage (termid.tsx): 
- useEffect(router.push(apiErrorPage), [error]);

pranavphadke1:
Need to do:
- [x] Rename page to "error"
- [x] Implement Vanessa's new Button design
- [x] Experiment with z-index to fix mobile page viewing size issue
- [x] Redirect to error page when searching with API down
- [x] Console log the specific error we run into

pranavphadke1:
These are the console log messages I added @hankewyczz 

When redirected to error page from the main page:
image

When redirected to error page from a failed search:
image

pages/error.tsx
```diff
@@ -0,0 +1,84 @@
...
+ * Page to indicate the api is down.
+ */
+export default function ApiErrorPage(): ReactElement {
+  const containerClassnames = 'home-container';
+
+  const [modalOpen, setModalOpen] = useState(false);
+
+  const toggleModal = () => {
+    setModalOpen(!modalOpen);
+  };
```
hankewyczz:
```suggestion
  const toggleModal = () => setModalOpen(!modalOpen);
```
a small nit

pages/error.tsx
```diff
@@ -0,0 +1,84 @@
...
+
+  return (
+    <>
```
hankewyczz: Can we not just use the top-level `div` here?

pages/error.tsx
```diff
@@ -0,0 +1,84 @@
...
+                    className="suport-button"
+                    onClick={toggleModal}
+                  >
+                    SEND SUPPORT REQUEST
+                  </button>
+                  <a
+                    target="_blank"
+                    rel="noopener noreferrer"
+                    className="create-issue"
+                    href="https://github.com/sandboxnu/searchneu"
```
hankewyczz:
```suggestion
                    href="https://github.com/sandboxnu/searchneu/issues/new"
```

styles/pages/_Error.scss
```diff
@@ -0,0 +1,158 @@
...
+  @media only screen and (min-width: 0px) {
+    // Screen too small to display
+    width: 0%;
+  }
+
+  @media only screen and (min-width: 200px) {
+    width: 45%;
+    right: 0px;
+    bottom: -10px;
+  }
```
hankewyczz: I'm not a fan of this pattern — I think it could be simplified. 

I think the ideal setup would be using flex columns — https://css-tricks.com/snippets/css/a-guide-to-flexbox/

When the screen is wide enough, it's a grid with two columns, one row (the husky in the rightmost cell)

When it isn't, it's a grid with one column and two rows (husky in the bottom cell).

Tbh, we could use flexbox for more of the elements here, like arranging the buttons. This shouldn't be something we have to create on our own.

hankewyczz (changes requested):

pranavphadke1:
@hankewyczz Redid the CSS using flexbox. This helped remove a lot of hard-coded styling and also got rid of the need for z-index.

Overall, reduced majority of @ media calls. However, they are still some necessary usage such as adding flex-wrap to the "error-container" class. Also needed to use it to handle the vertical orientation on mobile screens.

pranavphadke1:
@Lucas-Dunker ready for review

Lucas-Dunker (approved):
Looks good to me!

pages/[campus]/[termId].tsx
```diff
@@ -39,11 +38,12 @@ export default function Home(): ReactElement {
...
   const containerClassnames = 'home-container';
 
-  const [modalOpen, setModalOpen] = useState(false);
-
-  const toggleModal = () => {
-    setModalOpen(!modalOpen);
-  };
+  useEffect(() => {
+    if (termInfosError) {
+      console.log("We've encountered an error: " + termInfosError);
```
Comment (sebwittr) [inline pages/[campus]/[termId].tsx:43]: You should get rid of any console.logs you accidentally leave in.

Comment (pranavphadke1) [reply pages/[campus]/[termId].tsx:43]: These are to let users know what specific error occured (if they're curious). This was something Zachar asked to add. Is this how you want this implemented? @hankewyczz ?

pages/error.tsx
```diff
@@ -0,0 +1,87 @@
...
+                    Something went wrong...
+                  </div>
+
+                  <div className="api-error-text">
+                    {"Don't worry, we're on it"}
+                  </div>
+                  <div className="request-container">
+                    <button
+                      type="button"
+                      className="suport-button"
```
sebwittr: typo: probably should be "support button"

sebwittr (approved):
Code looks good! I noticed a typo in the code, but it doesn't matter too much.

pages/[campus]/[termId].tsx
Comment (sebwittr) [reply pages/[campus]/[termId].tsx:43]: ok cool sounds good then!

styles/pages/_Error.scss
```diff
@@ -0,0 +1,150 @@
+@import url('https://fonts.googleapis.com/css2?family=Montserrat+Alternates:wght@700&display=swap');
+
+/* Styles for ApiError Page. */
+.error-container {
+  height: 100vh;
+  width: 100vw;
+  display: flex !important;
```
hankewyczz: Try to make sure that all `!important` uses are actually needed — if not, we don't want to introduce too many (not saying this one is unnecessary, just asking for a second lookthrough)

styles/pages/_Error.scss
```diff
@@ -0,0 +1,150 @@
...
+}
+
+.error-text-container {
+  display: flex !important;
+  flex-direction: column;
+  justify-content: center;
+}
+
+.api-error-title-text {
+  font-family: Montserrat Alternates;
```
hankewyczz:
```suggestion
  font-family: 'Montserrat Alternates';
```

hankewyczz (approved):
Approved, but two quick things first:
- Try to remove any `!important` cases which aren't necessary
- Double-check with Vanessa (send her the staging link) that the design is OK

Looks great!

PR #208, opened 2023-04-19
Update Color Scheme To Sass Variables, Replace Sass @Import with @Use

Lucas-Dunker:
# Purpose

I TOLD YOU I WOULD DO IT TONIGHT!

This ticket is pretty self-explanatory; it refactors pretty much **every** color used in Search to a variable in-line with our color scheme (thanks, @vanessa-peng!) with the added bonus of changing our Sass use of @ import to @ use ([here's why](https://sass-lang.com/documentation/at-rules/import))

# Tickets

_What Trello tickets (if any) are associated with this PR?_

-[This one!](https://trello.com/c/o7uqfcwh/383-update-fe-color-scheme)

# Contributors

@Lucas-Dunker 

# Feature List

-Standardizes a SearchNEU color scheme for future use (as a result, some of the colors look different; @vanessa-peng please review!)

-Refactors most color hex codes to a color variable, both in `.tsx` files with the use of `_exports.module.scss` and in `.scss` files with the use of `_variables.csss`

-Changes our imports in Sass to use to protect from future deprecation

-Now that our colors can be changed from one location, we have the possibility to have themes or dark mode for Search! (as a feature down the line, of course)

# Pictures

NOTE: Some of the colors have been changed slightly to reflect our new standardized color palette? However, if anyone thinks these new colors don't work well, we can totally find a substitution for them or add our old colors back to the palette.

**Before is on the top, after is on the bottom.**

**Homepage:**
image
image

**Search Page:**
image
image

**Filters:**
image
image

**Class Page:**
image
image

Please check out the Vercel deployment as well! It's very likely I missed something on this huge refactor, so if everyone could set aside 30ish minutes to look at this in depth I'd really appreciate it! The error pages and feedback modals are mostly unchanged (aside from some small font color stuff)

# Reviewers

**Primary**:

@hankewyczz 
@sebwittr 
@vanessa-peng 

**Secondary**:

@pranavphadke1 
@soulwa 
@allychao


components/ClassPage/SectionsTermNav.tsx
```diff
@@ -32,7 +33,7 @@ export default function SectionsTermNav({
         }`}
       >
         <LeftNavArrow
-          fill={leftNavDisabled(currTermIndex) ? '#969696' : '#000000'}
```
Lucas-Dunker, line 35: Don't get scared at the total amount of files changed haha, most are simple color substitutions like this one.

components/ResultsPage/MobileSearchOverlay.tsx
```diff
@@ -4,6 +4,7 @@ import macros from '../macros';
 import FilterPanel from './FilterPanel';
 import FilterPills from './FilterPills';
 import { areFiltersSet, FilterOptions, FilterSelection } from './filters';
+import Colors from '../../styles/_exports.module.scss';
```
Lucas-Dunker, line 7: This is the import for using color variables in Typescript

components/icons/IconArrow.tsx
```diff
@@ -3,7 +3,7 @@ import React, { ReactElement } from 'react';
 const IconArrow = ({
   width = '11',
   height = '10',
-  fill = '#505050',
+  fill = '#4e525a',
```
Lucas-Dunker: A big part of this ticket was standardizing our color scheme (using [this](https://www.figma.com/file/ribLh5rcpGvejjQoTCODJX/%F0%9F%8E%A8-SearchNEU-Design-System---WIP?node-id=0-1) as a reference); additionally, I decided not to use color variables for SVG's, since I thought it wouldn't add much additional values compared to actual `tsx` or `sass` files

components/icons/IconUser.tsx
```diff
@@ -17,15 +17,15 @@ const IconUser = ({
     fill="none"
     xmlns="http://www.w3.org/2000/svg"
   >
-    <circle cx="14" cy="14" r="13" stroke="#A8DADC" strokeWidth="2" />
+    <circle cx="14" cy="14" r="13" stroke="#a8d8dc" strokeWidth="2" />
```
Lucas-Dunker, line 20: Another standardization was making uppercase hex codes lowercase!

generated/graphql.ts
```diff
@@ -177,6 +177,7 @@ export type SearchResultItem = ClassOccurrence | Employee;
 
 export type SearchResultItemConnection = {
   __typename?: 'SearchResultItemConnection';
+  totalCount: Scalars['Int'];
```
Lucas-Dunker: I believe this was because of a change from another ticket, but I'm not sure; @hankewyczz thoughts?

index.d.ts
```diff
@@ -7,3 +7,8 @@ declare module '*.yml' {
   const data: any;
   export default data;
 }
+
+declare module '*.scss' {
```
Lucas-Dunker, line 11: Adding `scss` files to the file types that typescript can read

jest.config.js
```diff
@@ -5,6 +5,6 @@ module.exports = {
     '^.+\\.yml$': '<rootDir>/node_modules/yaml-jest',
   },
   moduleNameMapper: {
-    '^.+\\.(css|less)$': '<rootDir>/__mocks__/style.js',
+    '^.+\\.(css|less|scss)$': '<rootDir>/__mocks__/style.js',
```
Lucas-Dunker, line 8: Letting our testing library understand `scss` files

styles/_AlertBanner.scss
```diff
@@ -1,11 +1,13 @@
+@use 'variables' as Colors;
```
Lucas-Dunker, line 1: This is the import for `scss` files!

styles/_variables.scss
```diff
@@ -4,3 +4,24 @@
  *
  * This file contains global variables, used across all styles.
  */
+
+//SearchNEU Color Palette
```
Lucas-Dunker, line 8: All of the colors SearchNEU now uses! (aside from the colors forming the gradient on the background of the homepage)

styles/_variables.scss
```diff
@@ -4,3 +4,24 @@
  *
  * This file contains global variables, used across all styles.
  */
+
```
Lucas-Dunker, line 7: All of the colors SearchNEU now uses! (aside from the colors forming the gradient on the background of the homepage)

styles/_variables.scss
```diff
@@ -4,3 +4,24 @@
...
+//Utility Colors
+$White: #fff;
+$Off_White: #f8f9f9;
+$Black: #000;
+$Grey: #6a7079;
+$Dark_Grey: #4e525a;
+$Light_Grey: #d1d3d7;
+$Green: #1f8b2a;
+$Light_Green: #a9dfae;
+$Light_Blue: #b5cad8;
```
Lucas-Dunker: All of the colors SearchNEU now uses! (aside from the colors forming the gradient on the background of the homepage)

components/ResultsPage/FeedbackModal/FeedbackModal.tsx
```diff
@@ -136,7 +136,7 @@ export default function FeedbackModal(): ReactElement {
             setOpen(!open);
           }}
         >
-          <LogoInput height="14" width="18" fill="#d41b2c" />
```
hankewyczz, line 139: Why was this change made & why doesn't it have a variable? (edit: isn't this `neu_red`?)

components/common/AlertBanner.tsx
```diff
@@ -40,7 +41,7 @@ export default function AlertBanner({ alertBannerData }: AlertBannerProps) {
             tabIndex={0}
             onClick={() => setIsVisible(false)}
           >
-            <IconClose fill="#808080" />
```
hankewyczz, line 43: Two issues:
- `Colors.gray` doesn't seem to exist? It's `grey`, not `gray`
- `Colors.grey` is a different shade of grey than `#6a7079`.

components/icons/Husky.tsx
```diff
@@ -19,7 +19,7 @@ export default function Husky({ className, campus }: HuskyProps): ReactElement {
       <g filter="url(#filter0_d)">
         <path
           d="M171.551 380.888s22.9-97.892 17.927-134.858c2.967-57.946-17.927-70.408-6.477-115.389 11.45-44.981-2.967-36.463 11.952-44.94 14.918-8.475 48.933 51.192 48.933 51.192s10.907 12.714 16.59 3.231c0 0 5.683 2.475 5.976 6.587.292 4.07 14.082-1.468 15.419-4.238 1.338-2.769 7.272-12.42-.877-53.876 0 0 .125-17.203 10.447-20.728 7.396-1.385-1.755-.294 13.08 9.19 14.834 9.482 40.158 36.001 47.262 59.624 0 0 18.679 12 20.935 17.539 0 0 6.937 1.217 10.364 3.105 3.468 1.93 8.483-9.902 24.738-13.175 16.256-3.273 29.377-14.77 46.51-6.084 12.787 17.203 3.301 19.091 3.301 19.091s.836 3.231 3.385 2.854c1.964 2.139 4.095 10.364 2.215 13.427-1.88 3.063 12.369 14.434 3.552 39.442-8.859 25.008-28.625 51.862-28.625 51.862s-10.238 21.441-6.143 38.477c4.096 17.036 23.618 68.796 17.015 78.279-6.602 9.441-65.656 69.125-114.506 66.146-48.892-2.979-155.66-13.217-162.973-66.758z"
-          fill="#fff"
```
hankewyczz: Why was this change made?

components/notifications/modal/SignUpModal.tsx
```diff
@@ -128,7 +129,7 @@ export default function SignUpModal({
       <div className="phone-modal">
         {isLoading && (
           <div className="phone-modal__spinner">
-            <MoonLoader color={'#7fc4c7'} loading={true} size={32} />
```
hankewyczz, line 131: Ditto here - is the new color what's in the Figma? These are the things that Vanessa should review - is the new usage of the changed color OK?

generated/graphql.ts
hankewyczz: This shouldn't be in your PR — run `git checkout origin/master -- generated/graphql.ts`

styles/_CheckboxFilter.scss
```diff
@@ -29,13 +31,13 @@
     border-radius: 3px;
 
     &:hover {
-      background-color: #eaeaea;
```
hankewyczz, line 32: Same here - I won't leave a comment in every spot, but these are the kinds of design changes that need to be reviewed. Going from `#FFFFFF` to `white` doesn't matter - going from a different color to white does matter

hankewyczz (changes requested):

sebwittr (approved):
There was one place that Zachar mentioned where a variable wasn't being used, but other than that everything looks good!

components/ResultsPage/FeedbackModal/FeedbackModal.tsx
Lucas-Dunker: I decided not to use color variables for SVGs and their corresponding tsx files, especially for fill attributes; I didn't think it made much semantic sense to make them variables since lots of them were likely imported straight from figma and they're more closely related to the actual shapes/components themselves than to styling (plus, there are lots of them). Let me know if this makes sense! If it doesn't, I can definitely go in and abstract them to variables

components/common/AlertBanner.tsx
Lucas-Dunker: 1. Whoops! Thanks for the catch
2. Yep, a big part of this ticket was standardizing the colors to our palette on the figma, so some of the colors changed slightly; this is intentional!

components/icons/Husky.tsx
Lucas-Dunker: that mightve been a strange autoformatter thing, ill go back and fix that

components/notifications/modal/SignUpModal.tsx
Lucas-Dunker: Yep, thats the new figma color, waiting for vanessa's feedback!

styles/_CheckboxFilter.scss
Lucas-Dunker: Agreed; I'll switch @vanessa-peng from a secondary reviewer to a primary one.

components/ResultsPage/FeedbackModal/FeedbackModal.tsx
hankewyczz: I think this would be a good spot to use the color variables. In static content (like the SVGs themselves), you're right; there's no point in using variables there, since those are directly exported from Figma. 

However, in a spot like this, where the content is actually being used (in a variable setting), we should use the color variables. (ig a simpler way of explaining it is that if a dev wouldn't ever actually type this, the hex code is OK. If a dev would write this, use the variables)

Lucas-Dunker:
Vanessa has approved the color changes!

hankewyczz (changes requested):
This looks great! I just have one suggestion — we need a way of enforcing this. New devs will onboard not knowing about this, and we should have guardrails (like the Github tests) to tell them that's not the right way of doing things. 

We can use the [`no-restricted-syntax`](https://eslint.org/docs/latest/rules/no-restricted-syntax)
rule in ESLint to ensure that our linter catches bad uses of hex codes. 

For our uses, something like this should work:
```js
"no-restricted-syntax": [
      "error",
      { "selector": "Literal[value=/^#[a-zA-Z0-9]/i]", "message": "Use 'Colors' from '<location>'" },
      { "selector": "Literal[value=/^rgb[(]/i]", "message": "Use 'Colors' from '<location>'" },
      { "selector": "Literal[value=/^rgba[(]/i]", "message": "Use 'Colors' from '<location>'" }
    ]
```

That rule will create a linter error every time someone tries using a hex code, RGB, or RGBA. It won't cover the case where someone uses a literal color (eg. `color: red`), but that's better than nothing!

You can disable this rule using [ESLint comments](https://eslint.org/docs/latest/use/configure/rules#disabling-rules) to exclude specific sections (like the SVG files).

Lucas-Dunker:
> This looks great! I just have one suggestion — we need a way of enforcing this. New devs will onboard not knowing about this, and we should have guardrails (like the Github tests) to tell them that's not the right way of doing things.
> 
> We can use the [`no-restricted-syntax`](https://eslint.org/docs/latest/rules/no-restricted-syntax) rule in ESLint to ensure that our linter catches bad uses of hex codes.
> 
> For our uses, something like this should work:
> 
> ```js
> "no-restricted-syntax": [
>       "error",
>       { "selector": "Literal[value=/^#[a-zA-Z0-9]/i]", "message": "Use 'Colors' from '<location>'" },
>       { "selector": "Literal[value=/^rgb[(]/i]", "message": "Use 'Colors' from '<location>'" },
>       { "selector": "Literal[value=/^rgba[(]/i]", "message": "Use 'Colors' from '<location>'" }
>     ]
> ```
> 
> That rule will create a linter error every time someone tries using a hex code, RGB, or RGBA. It won't cover the case where someone uses a literal color (eg. `color: red`), but that's better than nothing!
> 
> You can disable this rule using [ESLint comments](https://eslint.org/docs/latest/use/configure/rules#disabling-rules) to exclude specific sections (like the SVG files).

Agreed! I'll do this in a new PR for organization.

hankewyczz (approved):

PR #212, opened 2023-05-23
Add linting rules to error on non-variable color codes 

Lucas-Dunker:
# Purpose

Now that all of the dev-edited hex codes have been replaced with color variables, we should probably have some sort of linting rule that enforced them (as per @hankewyczz 's suggestion). This PR adds exactly that!

# Tickets

No associated trello ticket lol
image

# Contributors

@Lucas-Dunker 

# Feature List

-Linting rule for using raw hex codes, RGB, or RGBA values in Typescript files
-Addition of the `stylelint` package, setting up linting for raw hex codes, RGB, or RGBA values in SCSS files
-Some additional comments for using color hex codes in development
-Refactor a small handful of hex codes that were forgotten in the big refactor (thanks, linter!)

# Reviewers

**Primary**:
@hankewyczz @sebwit20 

**Secondary**:
@pranavphadke1 @soulwa


.eslintrc.json
```diff
@@ -23,7 +23,22 @@
   "rules": {
     "react/react-in-jsx-scope": "off",
     "@typescript-eslint/explicit-function-return-type": "off",
-    "@typescript-eslint/explicit-module-boundary-types": "off"
+    "@typescript-eslint/explicit-module-boundary-types": "off",
+    "no-restricted-syntax": [
```
hankewyczz, line 27: [nit] could you add some more detail into the message (or add a comment, but I'd prefer the former so they don't have to go looking for it)? Imagine you're a new developer who expects hex codes to work as usual - what's important for them to know?

package.json
```diff
@@ -95,6 +98,7 @@
     "jest": "^26.6.3",
     "next-compose-plugins": "^2.2.1",
     "next-test-api-route-handler": "^1.1.3",
+    "postcss": "8",
```
hankewyczz:
```suggestion
    "postcss": "^8",
```

pages/_document.tsx
```diff
@@ -2,6 +2,8 @@ import Document, { Head, Html, Main, NextScript } from 'next/document';
 import { ReactElement } from 'react';
 import Macros from '../components/abstractMacros';
 
+/* eslint-disable no-restricted-syntax */
```
hankewyczz, line 5: I think this is a good spot to use the variable instead of ignoring the lint rule

hankewyczz (changes requested):
Looks great! One big thing - could you break out the Node upgrade into a separate PR? It's easier to track/review that way (there's some parts of the codebase that need to also be upgraded [off the top of my head, `.nvmrc`])

pages/_document.tsx
Lucas-Dunker: Unfortunately document.tsx doesn't allow css, I tried for the last pr but the project wouldn't compile

.eslintrc.json
Lucas-Dunker: Totally!

Lucas-Dunker:
> Looks great! One big thing - could you break out the Node upgrade into a separate PR? It's easier to track/review that way (there's some parts of the codebase that need to also be upgraded [off the top of my head, `.nvmrc`])

I can definitely do that, but that PR will have to be approved first for this PR to work!

hankewyczz:
> I can definitely do that, but that PR will have to be approved first for this PR to work!

Yep, that's fine - I typically had a few PRs up like that at any given time. While you're working on the new branch (let's call it `node-upgrade` for this), you can just `pull origin/node-upgrade` into this branch. That'll ensure that this branch still works as intended.

I'd recommend adding `[DRAFT]` to the title of this PR until the node upgrade PR is merged

sebwittr (approved):
I think this is a good lint rule, just make sure to add a description to the rule so it's obvious why it's erroring.

hankewyczz (commented):
comment

hankewyczz (approved):

PR #213, opened 2023-05-24
Upgrade node versions in github workflows to match project node version

Lucas-Dunker:
# Purpose
 
While working on a linting ticket, I noticed that the node version used in our github workflows for testing, linting, etc. is a different version than the one in our `nvmrc` file. This naturally isn't a good idea, since we want our workflows to match our project environment as closely as possible.

# Tickets

No associated ticket 🙏 

# Contributors

@Lucas-Dunker 

# Feature List

-Update github workflow node version to match the node version in our `nvmrc` file

# Reviewers

**Primary**:
@hankewyczz


hankewyczz (changes requested):
pls also update `.nvmrc` and `package.json` to make sure our version is changed to the same thing across the board! Doing a project-wide search for `12` and `v12` ought to help with this (omit `.htm[l?],.md,.lock,.svg,.tsx,.js,.snap,.scss` - that should help). 

For `package.json`, we want to restrict the node version to 14+ - checkout CCA for an example of how to do that

Lucas-Dunker:
> pls also update `.nvmrc` and `package.json` to make sure our version is changed to the same thing across the board! Doing a project-wide search for `12` and `v12` ought to help with this (omit `.htm[l?],.md,.lock,.svg,.tsx,.js,.snap,.scss` - that should help).
> 
> For `package.json`, we want to restrict the node version to 14+ - checkout CCA for an example of how to do that

`nvmrc` is 14.19 as well so no issue there, but ill add an `engines` field to `package.json`!

hankewyczz (approved):

PR #215, opened 2023-09-16
Upgrade Node to Version 18 + Next to Version 12, React to Version 17.0.2

Lucas-Dunker:
# Purpose

This ticket was a trip. 

In short, while trying to do minor changes to the SearchNEU README,  I was getting errors about upgrading our node version:
image

If only changing our node version was that simple. Due to a bug in 2022-ish postCSS (and therefore all the libraries that use it) that made it incompatible with node versions 16 or greater, I had to upgrade many, many more things than anticipated. 

The good news: Search's dependencies are a lot newer!
The bad news: getting there was such a headache of trial and error that I never want to do it again,

# Contributors

@Lucas-Dunker 

# Feature List
* React upgraded to version 17.0.2
* Node upgraded to version 18
* NextJS upgraded to version 12
* Miscellaneous dependencies have been upgrade as needed
* Our yarn warnings on installation have been cut by over half!

# Reviewers

**Primary**:

@sebwit20


sebwittr (approved):
thank you for doing this!

PR #217, opened 2023-10-16
Set default dev environment variable

Lucas-Dunker:
# Purpose

When running SearchNEU's frontend locally, there's been a message for a while now saying something along the lines of "unknwon env! setting DEV". This ticket adds an environment variable to our dev config that explicitly set our environment to dev when running locally, getting rid of that console message.

# Tickets

[Trello Ticket](https://trello.com/c/e4blgbKp/404-unknown-env-when-running-locally)

# Contributors
@Lucas-Dunker 

# Feature List

* Add an explicit dev variable in `.env.development` rather than having a script in utils catch it for us. 

# Pictures

image

# Reviewers

Primary reviewer:

@sebwit20 

Secondary reviewers:

@pranavphadke1


pranavphadke1 (approved):

sebwittr (commented):
Is there a reason we're adding all these changes in linting to this PR? It seems similar to stuff that should've already been pushed to master

Lucas-Dunker:
> Is there a reason we're adding all these changes in linting to this PR? It seems similar to stuff that should've already been pushed to master

It just took all the commits from the rebase, none of its actually apart of the ticket

PR #230, opened 2023-11-05
Fix Missing "Prerequisites For"

### LINKED ISSUE #229: "Prerequisite For" field is inaccurate
I'm looking through courses for Spring 2024 and courses like CS 2500 list no courses that it is a prerequisite for (and there are _many_ considering that it's Fundies 1). 
image

Other examples (that I originally found this issue with) were CS 2810 and CS 2800.

Having this feature fixed would be incredibly useful for course planning if we're able to see the future courses that would be unlocked if I were to take this particular course.

AlpacaFur:
# Purpose

This fixes #229. I think the prereq/coreq access logic got copied when prereqFor has a different data format (flat list of classes).

# Tickets

N/A >:)

# Pictures
**Before: [link to prod](https://searchneu.com/NEU/202430/classPage/CS/2500)**
before-change

**After: [link to staging!](https://searchneu-git-luke-prereq-for-fix-sandboxneu.vercel.app/NEU/202430/classPage/CS/2500)**
after-change

# Reviewers
@Lucas-Dunker

### DIFF
--- a/components/ClassPage/ClassPageReqsBody.tsx
+++ b/components/ClassPage/ClassPageReqsBody.tsx
@@ -99,18 +99,16 @@ export default function ClassPageReqsBody({
           }
           className="prereqsFor"
           body={
-            latestOccurrence.prereqsFor.values.length === 0 ? (
+            latestOccurrence.prereqsFor.length === 0 ? (
               <span className="noReqs">None</span>
             ) : (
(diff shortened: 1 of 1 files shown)


Lucas-Dunker (approved):
THANK YOU LUKE!!

PR #238, opened 2024-01-22
Add ctrl+click to the "View more info for this class" button, remove "back to search results" url.

Lucas-Dunker:
# Purpose

This ticket turns the "View more info for this class' button into a url, allowing it to be more easily copied and ctrl-clicked. In the process, I found a bug related to graphql making queries before all the react data is loaded; a conditional has been added to wait for the necessary data to load before any data is fetched.

I also decided to remove the "back to search results" url, since I felt that it was unnecessary - other search engines like YouTube and Google don't have back buttons on their sites, they just rely on the browser. I can definitely roll this back if you think it's better to have it!

Feel free to check out the vercel deployment and see this change for yourself! 

# Tickets

[This ticket!](https://trello.com/c/pk772KOj/342-allow-ctrlclicking-of-class-pages)

# Contributors

@Lucas-Dunker 

# Pictures
image

image

# Reviewers

**Primary**:
@pranavphadke1 

**Secondary**:
@sebwittr @ananyaspatil

### DIFF
--- a/components/ResultsPage/Results/CourseResult.tsx
+++ b/components/ResultsPage/Results/CourseResult.tsx
@@ -58,7 +58,7 @@ export function CourseResult({
   // TODO (sam 2023-03-09): this is necessary because of `useShowAll`, which should likely not be coupled to courses.
   const sortedSections = useMemo(
     () => sortSections(course.sections, userInfo),
-    [course]
+    [course.sections, userInfo]
   );
   const { optionalDisplay } = useResultDetail(course);
 
@@ -124,17 +124,14 @@ export function CourseResult({
               )}
             </div>
             <div className="SearchResult__panel--right">
-              <div
-                onClick={() =>
-                  router.push(
-                    `/${campus}/${termId}/classPage/${course.subject}/${course.classId}`
-                  )
-                }
-              >
-                <div className="view-more-info-container">
-                  <IconNotepad className="notepad-icon" />
+              <div className="view-more-info-container">
+                <IconNotepad className="notepad-icon" />
+                <a
+                  className="view-more-info-link"
+                  href={`/${campus}/${termId}/classPage/${course.subject}/${course.classId}`}
+                >
                   <span>View more info for this class</span>
-                </div>
+                </a>
               </div>
 
               <SignUpForNotifications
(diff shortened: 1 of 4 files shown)


pranavphadke1 (approved):
goat

components/ResultsPage/Results/CourseResult.tsx
```diff
@@ -124,17 +124,14 @@ export function CourseResult({
...
-                  router.push(
-                    `/${campus}/${termId}/classPage/${course.subject}/${course.classId}`
-                  )
-                }
-              >
-                <div className="view-more-info-container">
-                  <IconNotepad className="notepad-icon" />
+              <div className="view-more-info-container">
+                <IconNotepad className="notepad-icon" />
+                <a
```
sebwittr, line 129: We should probably use <Link> instead of an <a> tag, since it works with the React router and gives us this functionality still. I think with Next it also prerenders the pages that are being linked, so theres little to no waiting.

pages/[campus]/[termId]/classPage/[subject]/[classId].tsx
```diff
@@ -72,8 +52,32 @@ export default function Page(): ReactElement {
...
+            classId: coreqVal.classId,
+          });
+        });
+        setClassPageInfo(classPage);
+        setCoreqInfo(coreqInfoArray);
+      }
+    };
     loadClassPageInfo();
-  }, [subject, classId]);
+  }, [subject, classId, router]);
```
Comment (sebwittr) [inline pages/[campus]/[termId]/classPage/[subject]/[classId].tsx:80]: Is router required here as a dependency?

sebwittr (changes requested):
Thanks for making this! Just had one minor thing, but everything looks good.

components/ResultsPage/Results/CourseResult.tsx
Lucas-Dunker: Sorry, what should we use instead? If you’re referring to a `Link` tag, unfortunately that doesn’t allow ctrl clicking.

pages/[campus]/[termId]/classPage/[subject]/[classId].tsx
Comment (Lucas-Dunker) [reply pages/[campus]/[termId]/classPage/[subject]/[classId].tsx:80]: Yes, we get graphql errors otherwise.

components/ResultsPage/Results/CourseResult.tsx
sebwittr: oops yea I was talking about Link. Apparently someone had a similar issue (https://github.com/vercel/next.js/issues/7218), and the solution was to just put the href in the Link tag. If that doesn't work I'd be ok with the plain a tag.

PR #239, opened 2024-01-28
fix applied filters overflow behavior on desktop

soulwa:
# Purpose

Quick styling change to make sure applied filters stack in their flexbox instead of just overflowing off of the screen on desktop. Additionally, applied filters which overflow will be scrollable until `767px`, any larger device will observe the new stacking behavior.

# Tickets

https://trello.com/c/6eKA6SLP/224-filters-go-off-screen-when-too-many-are-added

# Contributors

@soulwa 

# Pictures
before:
Screenshot 2024-01-28 at 16 43 45
after:
Screenshot 2024-01-28 at 16 44 35

### DIFF
--- a/styles/_FilterPills.scss
+++ b/styles/_FilterPills.scss
@@ -71,10 +71,8 @@
       right: 50%;
       position: absolute;
       content: '';
-      width: 14px;
-      /* x size */
-      height: 1.5px;
-      /* x thickness */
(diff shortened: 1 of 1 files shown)


pranavphadke1:
sick

pranavphadke1 (approved):

PR #248, opened 2024-02-03
Initial Subscription page skeleton

pranavphadke1:
# Purpose
This is the initial skeleton code for developing the subscription page.

Note: To run this locally you need to set up Twillio. Steps to do this are [here](https://apidocs.searchneu.com/#/notifications/notifications)

# Tickets

- https://trello.com/c/DnjL1887

# Contributors

@pranavphadke1 @sebwittr @allychao 

# Feature List

- Displays all subscribed courses and sections at 'http://localhost:5000/subscriptions'
- Also displays sections that users haven't subscribed to but can (sections with less than 5 seats remaining)
- Can toggle to subscribe or unsubscribe from a section
- Added a link under the user icon that takes you to the subscription page (code commented out until page is finalized)

# To Do 
- Unsubscribe button on top right of class card doesn't do anything yet
- Need to add the header bar
- Make the class cards collapse to hide section info
- Additional styling

# Pictures
image

# Reviewers

**Primary**:
@sebwittr @soulwa 

**Secondary**:
@ananyaspatil @Anzhuo-W 
@Lucas-Dunker if you wanna :D


components/Header.tsx
```diff
@@ -198,6 +202,9 @@ export default function Header({
                   <span className="user-menu__item" onClick={onSignOut}>
                     Sign out of {userInfo.phoneNumber}
                   </span>
+                  {/* <span className="user-menu__item" onClick={subscriptionPage}>
```
pranavphadke1, line 206: This is commented out until the page is finalized

components/ResultsPage/ResultsLoader.tsx
```diff
@@ -36,6 +36,10 @@ export const getGroupedByTimeOfDay = (times): DayjsTuple[] => {
     const dayIndices = Object.keys(times);
 
     for (const dayIndex of dayIndices) {
+      // I ran into some cases where dayIndex isn't an array causing the foreach to break below.
```
pranavphadke1: Some bug I saw that might be fixed. Just leaving this here incase the same issue occurs in later development.

pages/subscriptions.tsx
```diff
@@ -0,0 +1,163 @@
+import { ReactElement, useEffect, useState } from 'react';
+import { useRouter } from 'next/router';
+import { gqlClient } from '../utils/courseAPIClient';
+import { PacmanLoader } from 'react-spinners';
+import useUserInfo from '../utils/useUserInfo';
+import { SubscriptionCourse } from '../components/types';
+import { ClassCard } from '../components/SubscriptionsPage/ClassCard';
+
+export default function SubscriptionsPage(): ReactElement {
+  const [userInfo, isLoading, fetchUserInfo] = useUserInfo();
```
pranavphadke1: On a userInfo object, we store what courses and sections a user is subscribed to with list of courseIds and sectionIds. For this subscription page to work, we need to use these ids to find the associated courses and list the sections one can/have subscribed to as well.

pages/subscriptions.tsx
```diff
@@ -0,0 +1,163 @@
...
+
+    if (!userInfo && !isLoading) {
+      router.push('/');
+      return;
+    }
+
+    const classMapping = new Map<string, SubscriptionCourse>();
+
+    // This uses subscribed course ids to find the associated courses and their sections
+    const fetchCourseNotifs = async (): Promise<void> => {
```
pranavphadke1, line 30: This function looks through userInfo.courseIds to find courses that a user might be subscribed to.

pages/subscriptions.tsx
```diff
@@ -0,0 +1,163 @@
...
+          host: host,
+          name: result.classByHash.name,
+          lastUpdateTime: result.classByHash.lastUpdateTime,
+          sections: filteredSections,
+        });
+      }
+    };
+
+    // This uses subscribed section ids to find the associated section and the associated course for that section
+    const fetchSectionNotifs = async (): Promise<void> => {
```
pranavphadke1, line 73: This function looks through userInfo.sectionIds to find their associated courses. If those courses were not found by the function above (fetchCourseNotifs), it will add the new course to the list of classes

components/Header.tsx
```diff
@@ -67,6 +67,10 @@ export default function Header({
     );
   };
 
+  const subscriptionPage = (): void => {
```
sebwittr: Probably should comment this out to appease the linter

pages/subscriptions.tsx
sebwittr: can we rename isLoading to isUserInfoLoading or something similar? It makes it a little more clear

pages/subscriptions.tsx
```diff
@@ -0,0 +1,163 @@
...
+          result.classByHash.subject + result.classByHash.classId;
+
+        const subject = result.classByHash.subject;
+        const classId = result.classByHash.classId;
+        const host = result.classByHash.host;
+        const termId = result.classByHash.termId;
+
+        const filteredSections = result.classByHash.sections
+          .filter((s) => {
+            return s.seatsRemaining <= 5;
```
sebwittr, line 49: We filter the sections to only keep the ones who have <= 5 seats remaining, since those are the ones you can sign up for notifs.

pages/subscriptions.tsx
```diff
@@ -0,0 +1,163 @@
...
+          })
+          .map((s) => {
+            return {
+              ...s,
+              campusDescription: '',
```
sebwittr: remove campusDescription from here and types.ts since its not used anywhere

components/SubscriptionsPage/ClassCard.tsx
```diff
@@ -0,0 +1,104 @@
...
+  return (
+    <ClassCardWrapper
+      headerLeft={
+        <>
+          <span className="SearchResult__header--classTitle">
+            {course.subject} {course.classId}: {course.name}
+          </span>
+          <LastUpdated
+            host={course.host}
+            prettyUrl={''} //Ignore this for now, the current link we use is outdated anyway. Will need to be course.prettyUrl
```
sebwittr, line 62: This might be useless after the ticket removing the link from the class card

components/ResultsPage/ResultsLoader.tsx
```diff
@@ -36,6 +36,10 @@ export const getGroupedByTimeOfDay = (times): DayjsTuple[] => {
     const dayIndices = Object.keys(times);
 
     for (const dayIndex of dayIndices) {
+      // There are some cases where dayIndex isn't an array, causing the foreach call to break below.
+      if (!Array.isArray(times[dayIndex])) {
```
sebwittr, line 40: if times[dayIndex] isn't an array what actually is it, and what happens on the actual site?

components/SubscriptionsPage/ClassCard.tsx
```diff
@@ -0,0 +1,104 @@
+import { ReactElement } from 'react';
```
sebwittr, line 1: How much of this is the same as the class card from Search results, and what here is different? We might want to name this component something like SubscriptionClassCard to make it more specific.

pages/subscriptions.tsx
```diff
@@ -0,0 +1,163 @@
...
+        });
+        const courseCode =
+          result.sectionByHash.subject + result.sectionByHash.classId;
+
+        // If course has already been found in fetchCourseNotifs(), continue
+        if (!classMapping.has(courseCode)) {
+          const sectionHashSlice = sectionId.split('/');
+          const courseHash = sectionHashSlice.slice(0, -1).join('/');
+
+          const courseResult = await gqlClient.getCourseInfoByHash({
```
sebwittr, line 86: This is repeated logic from the other course function, can we generalize it in another function so we don't have duplicate code.

sebwittr (changes requested):
Left some comments about cleaning up the code, and I had some questions. It is looking really good though.

components/ResultsPage/ResultsLoader.tsx
pranavphadke1: times[dayindex] should be the format [obj,obj, ...] but for some reason there are instances where it is just a single obj, without being wrapped by a list. I'm not too sure why this happens as the flow to get all this logic is the exact same as the results loader (which doesn't have the issue). For now let's leave this filter check in and look into finding the root cause before we push the final changes for the page

components/types.ts
```diff
@@ -92,6 +101,21 @@ export interface Section {
...
+  meetings: Meeting[];
+  seatsRemaining: number;
+  seatsCapacity: number;
+  waitCapacity: number;
+  waitRemaining: number;
+  honors: boolean;
+  campus: string;
+  lastUpdateTime: number;
+};
+
```
soulwa: is there a need for this type? it doesn't seem to be used anywhere and seems to just pollute the types more since it only differs by one field from `Section`, so it seems we could just use a `Section` when we want to return/receive a `SubscriptionSection`.

components/SubscriptionsPage/ClassCard.tsx
soulwa: +1, I think it might be worth looking at what you can offload from this component onto the search result. in particular we don't really need the `ClassCardWrapperType` and `ClassCardWrapper` - those are the exact same as `SearchResultProps` and `SearchResult`. ultimately I'm not too hung up on this though, seems to make sense and works

pages/subscriptions.tsx
soulwa: +1

components/SubscriptionsPage/ClassCard.tsx
```diff
@@ -0,0 +1,104 @@
...
+          </span>
+          <LastUpdated
+            host={course.host}
+            prettyUrl={''} //Ignore this for now, the current link we use is outdated anyway. Will need to be course.prettyUrl
+            lastUpdateTime={course.lastUpdateTime}
+            className="SearchResult__header--sub"
+          />
+        </>
+      }
+      headerRight={<button>Unsubscribe</button>}
```
soulwa, line 68: unsubscribe functionality isn't working (yet), but it might be worth clarifying what this is intended to do. should it unsubscribe the user from all sections, and the course? or should it just unsubscribe from the course? currently, there's no visual on this component which indicates that you are subscribed to a course and not a section (like there is in the results page), which gives the user weird situations where it maybe seems like they should be unsubscribed, eg.

Screenshot 2024-02-15 at 18 07 44

pages/subscriptions.tsx
soulwa: +1

components/SubscriptionsPage/ClassCard.tsx
pranavphadke1: I agree. I plan on abstracting all of this later once more works has been merged into this page since we already have branches working off this right now.

PR #255, opened 2024-02-21
Remove host and prettyUrl from LastUpdated component

pranavphadke1:
# Purpose
- It's hard to tell that the hyperlink on the globe icon of a class card is actually a link. Also, the CRN links for sections of a class is the exact same link as the one on the globe. We decided to remove the link from the globe as it is redundent.

# Tickets
- https://trello.com/c/AzK2BWUo

# Contributors

@pranavphadke1 

# Pictures

image

# Reviewers

**Primary**:
@sebwittr 

**Secondary**:
@ananyaspatil 
@Anzhuo-W 
@soulwa


sebwittr (approved):
Looks good

PR #257, opened 2024-02-24
Pp/sign in feature

pranavphadke1:
# Purpose

Right now the flow to sign up for notifications is confusing. You have to find a class that you can sign up for notifications on and click the “Notify me when seats open” button. This doesn’t actually sign you up for open seats. All it does is let you verify your number and sign in, thus updating the UserInfo object. We want to remove this confusing button and let the login flow be through the actual sign in button.

# Tickets

- https://trello.com/c/0Kb7J6jd

# Contributors
@pranavphadke1 

# Feature List

- Added new sign in button to the header. Also added a ref to the drop down menu to allow closing the menu when clicking outside of scope
- Abstracted the components that wrap the sign in button and the user dropdown menu into a new component "DropdownMenuWrapper"
- Removed the old signin button from the course results card
- Notifications table always displays so users know which classes they can subscribe to
- Added a new sign in modal step and modernized the old sign in modals

# Pictures
image
image
image
image
image
image

# Reviewers

Primary reviewer:
@sebwittr
@soulwa 

Secondary reviewer:
@ananyaspatil 
@Anzhuo-W


pranavphadke1:
Still need to move repeated code for login/logout/userinfo logic into utils folder but this code is reviewable rn

components/Header.tsx
```diff
@@ -20,35 +26,112 @@ import {
...
 type HeaderProps = {
   router: NextRouter;
   title: string;
   searchData: SearchResult;
   termAndCampusToURL: (t: string, newCampus: string, query: string) => string;
-  userInfo: UserInfo;
-  onSignOut: () => void;
+};
+
+export const DropdownMenuWrapper = (): ReactElement => {
```
pranavphadke1: Exported wrapper for the dropdown.
- Abstracted out to reduce repeated code
- Exported as we are now useing it on the landing page

utils/useUserInfo.ts
```diff
@@ -1,18 +1,32 @@
...
+type useUserInfoReturn = {
+  userInfo: UserInfo | null;
+  isUserInfoLoading: boolean;
+  fetchUserInfo: () => void;
+  onSignOut: () => void;
+  onSignIn: (token: string) => void;
+};
 
-const useUserInfo = (): [UserInfo | null, boolean, () => void] => {
+// Custom hook to maintain all userInfo related utility functions
```
pranavphadke1, line 14: All twillio/userInfo related data/functions is wrapped in this custom hook
- reduces duplicate code and centralizes related logic

utils/useUserInfo.ts
```diff
@@ -37,7 +51,7 @@ const useUserInfo = (): [UserInfo | null, boolean, () => void] => {
     fetchData();
   }, [cookies]);
 
-  return [userInfo, isUserInfoLoading, fetchUserInfo];
+  return { userInfo, isUserInfoLoading, fetchUserInfo, onSignOut, onSignIn };
```
pranavphadke1, line 54: Instead of returning a list that enforces all data/functions to be instantiated on hook usage (and written in order), we can use an object. This allows you to define whatever data/functions you need to pull out during calls to this hook.

i.e. Say we only need access to userInfo and onSignout.
- Using a list we need to do : const [userInfo, isUserInfoLoading, fetchUserInfo, onSignOut, onSignIn] = useUserInfo()
- With an object we can do: const {userInfo, onSignOut} = useUserInfo()

pages/[campus]/[termId]/classPage/[subject]/[classId].tsx
```diff
@@ -49,30 +44,9 @@ export default function Page(): ReactElement {
...
-      axios
-        .get(
-          `${process.env.NEXT_PUBLIC_NOTIFS_ENDPOINT}/user/subscriptions/${token}`
-        )
-        .then(({ data }) => {
-          setUserInfo({ token, ...data });
-        });
-    }
-  };
-
```
Comment (pranavphadke1) [inline pages/[campus]/[termId]/classPage/[subject]/[classId].tsx:73]: Lines 10-13, 20, 52-73: This sort of logic is repeated. Instead moved into the custom hook userUserInfo()

pages/[campus]/[termId]/search/[query].tsx
```diff
@@ -54,33 +49,6 @@ export default function Results(): ReactElement | null {
...
-      axios
-        .get(
-          `${process.env.NEXT_PUBLIC_NOTIFS_ENDPOINT}/user/subscriptions/${token}`
-        )
-        .then(({ data }) => {
-          setUserInfo({ token, ...data });
-        });
-    }
-  };
-
```
Comment (pranavphadke1) [inline pages/[campus]/[termId]/search/[query].tsx:83]: Removed this for the same reason as mentioned before. All this logic has been moved to the custom hook userUserInfo()

utils/useUserInfo.ts
pranavphadke1: This additionally allows us to remove passing down functions fetchUserInfo, onSignOut, and onSignIn as props to components.

pages/[campus]/[termId].tsx
```diff
@@ -33,9 +31,6 @@ export default function Home(): ReactElement {
   const LATEST_TERM =
     termInfos[campus].length > 0 ? termInfos[campus][0]['value'] : '';
   const termId = (router.query.termId as string) || LATEST_TERM;
-
-  const alertBanners = Object.values(alertBannersData) as [AlertBannerData];
```
Comment (pranavphadke1) [inline pages/[campus]/[termId].tsx:37]: This isn't used anywhere.

components/notifications/SignUpForSectionNotifications.tsx
```diff
@@ -3,38 +3,28 @@
...
+type SignUpForSectionNotificationsProps = {
   course: Course;
   userInfo: UserInfo;
-  onSignIn: (token: string) => void;
   showNotificationSignup: boolean;
   fetchUserInfo: () => void;
 };
 
-export default function SignUpForNotifications({
+export default function SignUpForSectionNotifications({
```
pranavphadke1, line 19: This function no longer signs you up for notifications, it only displays the toggle for signing up for section notifications. This is because the process for signing up for notifications is now moved into the header. Thus, the file has been renamed and there is no need for the conditional rendering on line 63.

components/Header.tsx
```diff
@@ -20,35 +26,112 @@ import {
...
+  const [showModal, setShowModal] = useState(false);
+  const [showMenuDropdown, setShowMenuDropdown] = useState(false);
+  const dropdownRef = useRef(null);
```
pranavphadke1, line 64: Issue: The current dropdown menu doesn't close when you click anywhere else on the page. You have to click on the menu again to close it.

Solution: refs and listeners.
- ref's are used to interact with DOM elements. You can give components refs to know whether or not the component is being interacted with.

components/Header.tsx
```diff
@@ -20,35 +26,111 @@ import {
...
+  useEffect(() => {
+    fetchUserInfo();
+  });
+
+  useEffect(() => {
+    const handleCloseDropdown = (event: Event): void => {
+      if (dropdownRef.current && !dropdownRef.current.contains(event.target)) {
+        setShowMenuDropdown(false);
+      }
+    };
```
pranavphadke1, line 71: This function handles closing the dropdown menu. It takes in an event which in our case is the user clicking on the screen. Line 58 checks if the ref exists (it is initially undefined) and checks if the area that the ref targets (the dropdown menu in our case) contains the area that the event occured on (event.target is essentially the location on the screen the user clicked). If the area the user clicked is not in the area binded by the ref, the dropdown menu is closed.

components/Header.tsx
```diff
@@ -20,35 +26,111 @@ import {
...
+
+    document.addEventListener('mousedown', handleCloseDropdown);
```
pranavphadke1, line 73: This adds an event to the dom. Whenever the user clicks on the page, the handleCloseDropdown function is called.

components/Header.tsx
```diff
@@ -20,35 +26,111 @@ import {
...
+
+    return () => {
+      document.removeEventListener('mousedown', handleCloseDropdown);
+    };
```
pranavphadke1, line 77: We want to remove the event from the dom when this component (Header) is unmounted.

components/Header.tsx
```diff
@@ -20,35 +26,111 @@ import {
...
+  }, []);
```
pranavphadke1, line 78: The empty dependency array is used so that the listener on the dom is added when the component is mounted and removed when the component is unmounted

pranavphadke1:
In the process of refactoring and adding more design logic

components/notifications/modal/PhoneNumber.tsx
```diff
@@ -9,44 +10,51 @@ interface PhoneNumberProps {
...
 
-export default function PhonNumber({
+export default function PhoneNumber({
   setPhoneNumber,
   onCancel,
   onSubmit,
   error,
 }: PhoneNumberProps): ReactElement {
+  const validatePhoneNumber = (number: string | undefined): void => {
+    if (typeof number === 'string') setPhoneNumber(number);
```
soulwa, line 20: should we be doing more validation here?

soulwa (approved):

soulwa:
generally looks good to me- no more need to copy paste user state functions everywhere. `onSignIn` now gets passed down through components which seems a bit messy but much less code smell vs. before. would also rather get this PR in vs. worrying about that since things are functional and fairly easy to understand

components/notifications/modal/GoSignIn.tsx
```diff
@@ -0,0 +1,46 @@
...
+import React, { ReactElement } from 'react';
+import X from '../../icons/X.svg';
+import OneMoreStep from '../../icons/one-more-step.svg';
+
+interface GoSignInProps {
+  onCancel: () => void;
+  onSubmit: () => void;
+}
+
+export default function GoSignIn({
```
soulwa, line 11: what's the user flow like to actually reach this code? running this locally and using the sign up button, I get the same flow that we usually have

soulwa (approved):

soulwa:
generally looks good to me- no more need to copy paste user state functions everywhere. onSignIn now gets passed down through components which seems a bit messy but much less code smell vs. before. would also rather get this PR in vs. worrying about that since things are functional and fairly easy to understand

also it's so weird i can approve twice lol. was just trying to add a comment to previous review..

components/notifications/modal/GoSignIn.tsx
pranavphadke1: We now always show the notifications table for sections so users know what sections they can sign up for. If they click on a toggle for a section that allows notifications, they are prompted to the "GoSignIn" modal

components/notifications/modal/PhoneNumber.tsx
pranavphadke1: There's some weird type error that currently exists in prod. If you type characters other than digits 0-9, the "number" passed to setPhoneNumber is undefined. This just makes sure that "phoneNumber" is always a string.

components/Header.tsx
SimChadha: Effect is only run on first render. Consider adding dropdownRef (or dropdownRef.current ?) to the list of dependencies so its run on ref change as well as first render.

useEffect(() => {
    const handleCloseDropdown = (event: Event): void => {
      if (dropdownRef.current && !dropdownRef.current.contains(event.target)) {
        setShowMenuDropdown(false);
        ...................
      }, **_[dropdownRef]};_**

components/Modal.tsx
```diff
@@ -23,16 +23,14 @@ export default function Modal({
...
           </div>
-        </div>
-      )}
-    </>,
-    document.body
-  );
+        </div>,
+        document.body
+      )
+    : null;
```
pranavphadke1, line 35: Needs a fallback here, otherwise causes an error locally

components/ResultsPage/Results/CourseResult.tsx
```diff
@@ -163,7 +157,7 @@ export function CourseResult({
                 <th> Meetings </th>
                 <th> Campus </th>
                 <th> Seats </th>
-                {userInfo && <th> Notifications </th>}
+                <th> Notifications </th>
```
pranavphadke1, line 160: We are now always showing notifications so that users can see what sections they can get notifications for

components/ResultsPage/ToggleFilter.tsx
```diff
@@ -25,7 +26,6 @@ export default function ToggleFilter({
           type="checkbox"
           onChange={onChange}
           id={id}
-          defaultChecked={selected}
```
pranavphadke1, line 28: Shouldn't have defaultChecked and checked

components/SubscriptionsPage/ClassCard.tsx
```diff
@@ -66,52 +71,54 @@ export const ClassCard = ({
...
-                  {sectionsFormatted.map((section) => (
-                    <DesktopSectionPanel
-                      key={section.crn}
-                      section={section}
-                      userInfo={userInfo}
-                      fetchUserInfo={fetchUserInfo}
-                    />
-                  ))}
-                </tbody>
+          <div style={{ display: areSectionsHidden ? 'none' : 'block' }}>
```
pranavphadke1, line 74: Instead of rerendering, we just want to hide the component while keeping it on the DOM. This keeps data consistent and reduces rerenders.

components/SubscriptionsPage/ClassCard.tsx
```diff
@@ -66,52 +71,54 @@ export const ClassCard = ({
...
+              </thead>
+              <tbody>
+                {sectionsFormatted.map((section) => (
+                  <DesktopSectionPanel
+                    key={section.crn}
+                    section={section}
+                    userInfo={userInfo}
+                    fetchUserInfo={fetchUserInfo}
+                    // We don't really have access to onSignIn until header is added to the subscription page. Passing in a fake onSignIn
+                    onSignIn={onSignIn}
```
pranavphadke1, line 101: Once header is added to the subscription page, will take a look at this onSignIn function... it needs to be synced to the header

PR #262, opened 2024-04-08
Pp/we need your help modal

pranavphadke1:
# Purpose

Created modal to collect user feedback on the impact of SearchNEU. Also created a toast wrapper component that can be used for any future toast designs. Removed the old feedback modal that showed on the bottom right of the results page.

# Contributors
@pranavphadke1 

# Feature List
- New modal on splash page
- Modal button redirects to a google form to collect feedback
- Modal shows once when user visits splash page. Stores a cookie so it isn't displayed again.
- Feedback toast shows on splash page and results page

# Pictures

image
image
image

# Reviewers

**Primary**:
@sebwittr

**Secondary**:
@Anzhuo-W @ananyaspatil


pranavphadke1:
Button on splash page will be added soon. Still finalizing design.

components/Modal.tsx
```diff
@@ -23,14 +22,11 @@ export default function Modal({
     return () => document.removeEventListener('keydown', handleKeyEvent);
   }, [onCancel]);
 
-  return visible
-    ? createPortal(
```
pranavphadke1, line 27: createPortal isn't needed here and causes bugs locally

pages/[campus]/[termId].tsx
```diff
@@ -35,7 +37,21 @@ export default function Home(): ReactElement {
 
   const { userInfo, fetchUserInfo, onSignIn, onSignOut } = useUserInfo();
 
+  const [showHelpModal, setShowHelpModal] = useState(true);
+
+  const fetchFeedbackToken = async (): Promise<void> => {
```
Comment (pranavphadke1) [inline pages/[campus]/[termId].tsx:42]: This checks if there is an existing cookie 'FeedbackModal JWT' and adds one if there isn't. This is because if the user has already seen the feedback modal, we don't want to keep displaying it to them.

utils/useUserInfo.ts
```diff
@@ -13,7 +13,7 @@ type useUserInfoReturn = {
 
 // Custom hook to maintain all userInfo related utility functions
 const useUserInfo = (): useUserInfoReturn => {
-  const [cookies, setCookies] = useState(new Cookies());
+  const cookies = new Cookies();
```
pranavphadke1, line 16: don't need useState here

sebwittr:
Minor bug, but everytime you refresh the page the modal flashes on screen and disappears. We probably want the modal to not be seen by default, and only show it if the cookie is enabled. Also are we putting the testimonial link anywhere else?

sebwittr (approved):
Ok I'm not getting the bug anymore, so this looks good. We can probably merge this right now.

ananyaspatil (approved):
Looks good!

components/Toast.tsx
```diff
@@ -0,0 +1,42 @@
+import { ReactElement, useEffect, useState } from 'react';
```
pranavphadke1, line 1: This is a wrapper toast component that can be used for future toasts.

PR #263, opened 2024-04-11
Add header to subscription page

soulwa:
# Purpose
adds the header to the subscription page. initially was doing more refactoring since the header is pretty tightly coupled to search functionality, but just did something to get it working for now since flexbox is a PITA

should be pretty straightforward, whoever wants to get eyes on it and verify behavior is good should

# Tickets

https://trello.com/c/PyoourKh/255-add-header-navbar-to-the-top-of-subscription-page

# Contributors

@soulwa 

# Pictures

before:

Screenshot 2024-04-11 at 12 07 11

now:
image

# Reviewers

Primary reviewers:
- @pranavphadke1 
- @sebwittr 

Secondary reviewers:
- @ananyaspatil 
- @Anzhuo-W

### DIFF
--- a/components/Header.tsx
+++ b/components/Header.tsx
@@ -191,11 +191,9 @@ export default function Header({
   };
 
   let backlink = '/';
-  if (campus != null) {
-    backlink += `${campus}/`;
-  }
+
   if (campus != null && termId != null) {
-    backlink += `${termId}/`;
+    backlink += `${campus}/${termId}/`;
   }
 
   if (showOverlay && macros.isMobile) {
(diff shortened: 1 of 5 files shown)


soulwa:
note: tests are broken since I changed the header structure- this is expected. however, they're also broken since they pass a searchData of `null` or `undefined` and expect the search bar to still be present. I'm now relying on that value being `null` to mean that we shouldn't render the search bar. would it be better if i just had a bool in the header for when we want to render the search bar?

pranavphadke1 (approved):

PR #265, opened 2024-04-14
Ap/no notifications husky page

ananyaspatil:
# Purpose
Creates an EmptyCard to display for the Empty Notifications page. 
When "Search for classes" button is hovered over, crying-husky image changes to happy-husky image. 
Includes a placeholder text for "Spring 2023 Notifications", this will need to be changed to actually reflect the semester of notifications being displayed.

# Tickets
https://trello.com/c/SoJpDQFG

-

# Contributors

@ananyaspatil @Anzhuo-W 

# Feature List
Created display for the Empty Notifications page. 
 - Created EmptyCard Component: contains display text and a button that redirects to the default search page
 - added crying-husky.svg and happy-husky.svg images
When "Search for classes" button is hovered over, crying-husky image changes to happy-husky image. 
 - Added isHovering const to EmptyCard
 - Added logic to update husky image to happy when OnMouseEnter is true for "search for classes" button
- Heading text updates to reflect the semester of notifications being displayed (should always be the most recent semester).

# Pictures
No Hover:
image

Hover:
image

# Reviewers

Primary reviewer:
@pranavphadke1 @sebwittr 

Secondary reviewers:
@soulwa


components/SubscriptionsPage/EmptyCard.tsx
```diff
@@ -0,0 +1,68 @@
...
+                    </div>
+                  </div>
+                }
+                headerRight={
+                  <div
+                    className="Results_Main__EmptyCard_Divider"
+                    onClick={() => {
+                      router.push('NEU');
+                    }}
+                    onMouseEnter={() => setIsHovering(true)}
```
ananyaspatil, line 58: To change image upon hovering, am setting isHovering constant to true when mouse enters the Button and to false when mouse leaves the button.

components/SubscriptionsPage/EmptyCard.tsx
```diff
@@ -0,0 +1,68 @@
...
+import React, { useState } from 'react';
+import { ClassCardWrapper } from './ClassCard';
+import { useRouter } from 'next/router';
+import Circular from '../icons/circular.svg';
+import CryingHusky from '../icons/crying-husky.svg';
+import HappyHusky from '../icons/happy-husky.svg';
+
+export const EmptyCard = (): ReactElement => {
+  const router = useRouter();
+  const [isHovering, setIsHovering] = useState(false);
```
ananyaspatil, line 13: Added const isHovering so that Husky image can be changed when mouse hovers over the "search for classes" button

pages/subscriptions.tsx
```diff
@@ -131,6 +135,10 @@ export default function SubscriptionsPage(): ReactElement {
         await fetchSectionNotifs();
         setClasses(classMapping);
         setIsFetching(false);
+        // are there classes the user is subscribed to?
```
ananyaspatil, line 137: Adding condition for when to display the No Notifications page

components/SubscriptionsPage/ClassCard.tsx
```diff
@@ -13,18 +13,17 @@ type ClassCardWrapperType = {
...
   body,
   afterBody,
 }: ClassCardWrapperType): ReactElement => {
   return (
     <div className="SearchResult">
-      <div className="SearchResult__header">
-        <div className="SearchResult__header--left">{headerLeft}</div>
-        {headerRight}
-      </div>
+      <div>{headerLeft}</div>
```
sebwittr, line 25: I think we needed the styles from SearchResult__header and SearchResult__header--left for the class cards. Did your changes actually need to change this file?

components/SubscriptionsPage/EmptyCard.tsx
```diff
@@ -0,0 +1,68 @@
...
+
+  return (
+    <>
+      <div className="Results_Container">
```
sebwittr: We probably don't need all these extra styles, and it would be better to make a separate style sheet for EmptyCards.

styles/pages/_Results.scss
```diff
@@ -154,6 +152,133 @@ $SIDEBAR_WIDTH: 268px;
...
+      line-height: 10px;
+      padding: 12px 20px;
+      align-items: center;
+    }
+
+    .Results_Main__EmptyCard_Divider {
+      padding-bottom: 0px;
+    }
+
+    .SearchResult {
```
sebwittr, line 221: I would move these out into a separate _EmptyCard.scss so we don't put too much into this file.

sebwittr (changes requested):
This looks good, main thing is the changes to ClassCardWrapper. I think we shouldn't need to touch that file at all, since it's only supposed to wrap actual class cards. The EmptyCard should serve as its own component, and not really need to rely on that.

Anzhuo-W:
Trello ticket

Screen Shot 2024-11-17 at 4 51 17 PM

cherman23:
Styles have been moved to new EmptyStyles, Empty Card automatically updates to show current semester.

components/SubscriptionsPage/ClassCard.tsx
```diff
@@ -13,18 +13,17 @@ type ClassCardWrapperType = {
   afterBody?: ReactElement;
 };
 
-const ClassCardWrapper = ({
+export const ClassCardWrapper = ({
```
mehallhm, line 17: Let's match the rest of the codebase here and use `function` definitions rather than arrow notations. It seems to be like a 70 / 30 split on what we do, but we should pick one to stick with, and I'll just go with the majority

package.json
```diff
@@ -43,6 +43,7 @@
     "js-yaml-loader": "^1.2.2",
     "jsonwebtoken": "^8.5.1",
     "jwt-decode": "^3.1.2",
+    "leading-trim": "^1.0.2",
```
mehallhm: I would like to see evidence that adding a package is the correct call here. It's fine if we actually need it, but I would like that proof

pages/subscriptions.tsx
```diff
@@ -133,6 +137,10 @@ export default function SubscriptionsPage(): ReactElement {
...
+        if (classMapping.size > 0) {
+          setIsSubscribed(true);
```
mehallhm, line 139: Why is a state variable being reassigned to another state variable? I would just use `classMapping.size > 0` instead of managing a derived piece of state

pages/subscriptions.tsx
```diff
@@ -158,25 +166,33 @@ export default function SubscriptionsPage(): ReactElement {
       {isFetching ? (
```
mehallhm: I don't love the ternary, especially nested ones. Would an early return be better? This is really not that important but just thinking from a readability standpoint.

mehallhm (commented):
Generally looks pretty good, just some comments on code consistency and readability for the rest of the codebase

pages/subscriptions.tsx
cherman23: we will address this in a future ticket

package.json
cherman23: addressed in recent commit

components/SubscriptionsPage/ClassCard.tsx
cherman23: addressed in recent commit

pages/subscriptions.tsx
cherman23: we will address this in a future ticket

PR #276, opened 2024-08-12
Ap/total subscriptions limit

ananyaspatil:
# Purpose
We are limiting users to 12 total subscriptions per semester to ensure SearchNEU can stay up and running for our growing user base. This implements these new sub limits and enables the sign in / notification buttons. A new modal is also added to update visitors on these new changes.

# Tickets
https://trello.com/c/JbNEe9f9/434-implement-12-total-subscriptions-limit-per-user

# Contributors
@ananyaspatil 

# Feature List
- Limit User to 12 total subscriptions (can be any combination of course or section subscriptions)
          ->  notification buttons are disabled when user has reached 12 subscriptions
- Modal w/ new limits info to replace current "Notifications Closed for the Summer" modal
- Enable 'Sign in' button

# Pictures

new modal:
image

# Reviewers
@pranavphadke1 @sebwittr @Lucas-Dunker @Anzhuo-W

### DIFF
--- a/components/ResultsPage/Results/NotifSignUpButton.tsx
+++ b/components/ResultsPage/Results/NotifSignUpButton.tsx
@@ -13,14 +13,16 @@ export default function NotifSignUpButton({
     macros.logAmplitudeEvent('Notifs Button');
   };
 
-  const NOTIFICATIONS_ARE_DISABLED = true;
+  const NOTIFICATIONS_ARE_DISABLED = false;
 
   return (
     <button
       disabled={NOTIFICATIONS_ARE_DISABLED}
       onClick={onClickWithAmplitudeHook}
       type="button"
-      className={`Results_SignIn ${NOTIFICATIONS_ARE_DISABLED && 'disabledButton'}`}
+      className={`Results_SignIn ${
+        NOTIFICATIONS_ARE_DISABLED ? 'disabledButton' : ''
+      }`}
     >
       Sign In
     </button>
--- a/components/panels/CourseCheckBox.tsx
+++ b/components/panels/CourseCheckBox.tsx
@@ -85,7 +85,6 @@ export default function CourseCheckBox({
       <div className="signUpSwitch toggle">
         <div className="notifSwitch">
           <input
-            //disabled={NOTIFICATIONS_ARE_DISABLED && (userInfo.courseIds.length + userInfo.sectionIds.length >= NOTIFICATIONS_LIMIT)}
             disabled={notificationsLimitReached()}
             checked={checked}
             onChange={onCheckboxClick}
--- a/components/panels/CourseCheckBox.tsx
+++ b/components/panels/CourseCheckBox.tsx
@@ -27,7 +27,15 @@ export default function CourseCheckBox({
 }: CourseCheckBoxProps): ReactElement {
   const [showModal, setShowModal] = useState(false);
   const [notifSwitchId] = useState(uniqueId('notifSwitch-'));
-  const NOTIFICATIONS_ARE_DISABLED = true;
+
+  const NOTIFICATIONS_LIMIT = 12;
+  const NOTIFICATIONS_ARE_DISABLED = false;
+
+  const notificationsLimitReached = (): boolean =>
+    NOTIFICATIONS_ARE_DISABLED ||
+    (userInfo &&
+      userInfo.courseIds.length + userInfo.sectionIds.length >=
+        NOTIFICATIONS_LIMIT);
 
   const isCourseChecked = (): boolean =>
     userInfo && userInfo.courseIds.includes(Keys.getClassHash(course));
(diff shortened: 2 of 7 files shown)


components/panels/CourseCheckBox.tsx
```diff
@@ -27,7 +27,15 @@ export default function CourseCheckBox({
...
   const [notifSwitchId] = useState(uniqueId('notifSwitch-'));
-  const NOTIFICATIONS_ARE_DISABLED = true;
+
+  const NOTIFICATIONS_LIMIT = 12;
+  const NOTIFICATIONS_ARE_DISABLED = false;
+
+  const notificationsLimitReached = (): boolean =>
+    NOTIFICATIONS_ARE_DISABLED ||
+    (userInfo &&
+      userInfo.courseIds.length + userInfo.sectionIds.length >=
```
Lucas-Dunker, line 37: So just confirm, section Ids and course Ids are stored separately for notification purposes?

Lucas-Dunker (approved):
Looks good! I can't test out the notif limits in staging, but Im happy for us to merge and try this out in production. How have you all been testing out the Twilio stuff locally?

components/panels/CourseCheckBox.tsx
ananyaspatil: Yes, they're stored separately since there's 2 types of notifs being sent out. courseIDs is to track new sections added to a course while sectionIds is to track new seats added to a section.

ananyaspatil: We'd been running the notifications server locally to test w/ twilio. I made this PR as a draft while I try to still test w/ twilio before publishing

Lucas-Dunker: ah gotcha, sounds good! ill re-mark this as a draft then

components/panels/CourseCheckBox.tsx
```diff
@@ -77,16 +85,23 @@ export default function CourseCheckBox({
       <div className="signUpSwitch toggle">
         <div className="notifSwitch">
           <input
-            disabled={NOTIFICATIONS_ARE_DISABLED}
+            //disabled={NOTIFICATIONS_ARE_DISABLED && (userInfo.courseIds.length + userInfo.sectionIds.length >= NOTIFICATIONS_LIMIT)}
```
pranavphadke1: remove comment

components/panels/CourseCheckBox.tsx
```diff
@@ -77,16 +85,23 @@ export default function CourseCheckBox({
...
             checked={checked}
             onChange={onCheckboxClick}
             className="react-switch-checkbox"
             id={notifSwitchId}
             type="checkbox"
           />
           <label
-            className={`react-switch-label ${NOTIFICATIONS_ARE_DISABLED && 'disabledButton'}`}
-            style={{ marginTop: '0px', cursor: `${NOTIFICATIONS_ARE_DISABLED ? 'not-allowed' : 'inherit'}` }}
+            //className={`react-switch-label ${NOTIFICATIONS_ARE_DISABLED && 'disabledButton'}`}
```
pranavphadke1, line 96: remove comment

pranavphadke1 (approved):
lgtm

sebwittr:
These changes look good, we just have to enforce it in the backend now too.

Anzhuo-W:
The changes look good! One thing is just the wording of the modal, it sort of makes me think that it's 12 ever, but it looks like the code's checking the list of subscriptions from userInfo, so we might want to add "concurrent" or something (& use the constant for the count). 

Also curious if we have the data for how many different notifications each user subscribes to per semester

ananyaspatil:
> The changes look good! One thing is just the wording of the modal, it sort of makes me think that it's 12 ever, but it looks like the code's checking the list of subscriptions from userInfo, so we might want to add "concurrent" or something (& use the constant for the count).
> 
> Also curious if we have the data for how many different notifications each user subscribes to per semester

Oh good point, I have updated the wording to "We're limiting users to 12 subscriptions at a time to keep SearchNEU running for our growing NEU community.". 
I don't believe we track a user's total different subscriptions in a semester. But, we do track the number of notifications users are subscribed to at a given moment(which, if we look at the end of a semester, would pretty much equal a user's total subscriptions over the semester bc they don't really unsubscribe rn).

Anzhuo-W (approved):
Looks good to me!

Anzhuo-W:
> But, we do track the number of notifications users are subscribed to at a given moment(which, if we look at the end of a semester, would pretty much equal a user's total subscriptions over the semester bc they don't really unsubscribe rn).

Ooh nice! Can we calculate what the mean/mode is for number of subscriptions? That could add to our value prop & proof of usage and maybe we choose 60th percentile for cutoff as a reason why the limit or sth

components/ResultsPage/Results/NotifSignUpButton.tsx
ananyaspatil, line 1: make NOTIFICATIONS_ARE_DISABLED a env variable

Issue #278, opened 2024-10-21
Signing in spams requests for the user's subscriptions


mehallhm:
Upon signing in, the application sends many requests to `https://stagingnotifs.searchneu.com/user/subscriptions/{token}` constantly, non-stop, on the pages for course search (home page, results page, ie NOT 404 page). As a result of these requests, Chromium based browsers experience large slowdowns and WebKit based browsers freeze and crash (Firefox impact unknown).

PR #281, opened 2024-10-30
Fixes notification spam upon sign in.

cherman23:
# Purpose

This fixes the spamming of requests upon user sign in. This also fixes the dropdown sign out menu that wouldn't load.

# Tickets

Github issue here : https://github.com/sandboxnu/searchneu/issues/278

# Contributors

@mehallhm 
@cherman23 

# Feature List

- The getUserInfo() method in the utils/useUserInfo updates a user's cookies on sign-in. Within the same method, this would cause useEffect() to trigger due to the cookies being changed. The useEffectHook would also update cookies, causing a positive feedback loop that would result in network spamming.
- The spamming is prevented by not updating cookies within the useEffectHook().
- This change comes from a previous PR, where cookies were changed from a state variable. This can be found here: https://github.com/sandboxnu/searchneu/commit/88b525a4b058aae6a3e0f16c486e8db9d081862e#diff-f733484387256e891c78b27d74d50b723f95c02dce0d5463b163b1a40d2155a6. Due to how react deals with state, the useEffect() was not triggered previously.

- The showMenuDropdown was initially set up using a useRef hook, which would not be tracked during rendering. This is fixed by changing the showDropDown to a state variable where it is rendered correctly.

# Reviewers

**Primary**:

@ananyaspatil 

**Secondary**:

@nickpfeiffer05 
@Anzhuo-W 
@olivial215 
@serena-boyu

### DIFF
--- a/components/Header.tsx
+++ b/components/Header.tsx
@@ -70,16 +70,17 @@ export const DropdownMenuWrapper = ({
   onSignIn,
 }: DropdownMenuWrapperProps): ReactElement => {
   const [showModal, setShowModal] = useState(false);
-  const showMenuDropdown = useRef(false);
+  // const showMenuDropdown = useRef(false);
+  const [showMenuDropdown, setShowMenuDropdown] = useState(false);
   const dropdownRef = useRef(null);
   const router = useRouter();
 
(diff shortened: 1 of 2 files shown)


ananyaspatil (approved):
looks good to me

PR #283, opened 2024-11-04
Allow newer versions of Node

mehallhm:
# Purpose

Removing the major version lock for Node while we migrate the backend, ensuing that one Node version can be used for both the frontend and backend.

# Tickets

Just following the [course-catalog](https://github.com/sandboxnu/course-catalog-api/tree/master) to ensure tooling is the same across both.

# Feature List
- Changes the engine version lock from Node 18 only to any version including or newer than Node 18

# Reviewers

**Primary**:
@ananyaspatil

**Secondary**:
@cherman23 @nickpfeiffer05  @Anzhuo-W


ananyaspatil (approved):
lgtm

PR #284, opened 2024-11-07
Remove subscription modal

cherman23:
# Purpose

After two weeks of turning on notifications modal, we are removing it from popping up. 
Removed instance of Notification from home page and update snapshots to match changes.

# Tickets

https://github.com/orgs/sandboxnu/projects/20?pane=issue&itemId=85788239

# Pictures

Before:
image

After:
image

# Reviewers

Primary reviewer:
@ananyaspatil 

Secondary reviewers:

@mehallhm 
@Anzhuo-W 
@nickpfeiffer05


mehallhm (commented):
I think you could go through and remove some of the cookie storage stuff too for the modal... like lines 54 through 87 or so in the [campus] [termId] page. I can't remember is more stuff scattered about, prob safe to just remove it now

mehallhm:
Closing b/c the modal updated and changed. Will file a new PR when the new modal should be removed

PR #286, opened 2024-11-19
Unable to disable notifications when the limit is reached

mehallhm:
# Purpose
Fixes a bug that too many people reported and I should have been in class

# Tickets
- No ticket was made even though there should have been

# Feature List
- Fixed the notification toggle locking after reaching the notification limit, preventing unsubscribing from notifs
- Add a new tootip that indicates the notification limit is reached to the toggle when it is locked

# Pictures
image
Notice the toggled toggle is still enabled and toggled while the disabled toggle is untoggled and untogglable. The third toggle is a visual bug that would take me longer than I care to commit right now but is also untogglable even though the toggle looks like it could be toggled. The tooltip was hard to show - trust me bro.

# Reviewers
**Primary**:
@ananyaspatil 

**Secondary**:
@nickpfeiffer05 @Anzhuo-W @cherman23


ananyaspatil (changes requested):
Changes look good to me! But believe these will also need to be added to CourseCheckBox.tsx

ananyaspatil (approved):

PR #330, opened 2025-03-23
Got rid of html tags before description is put into a paragraph tag

WesleyTran0:
# Purpose

Adds a strip html tag function
gets rid of the unwanted tags in the course description

# Tickets
#239 
-

# Contributors

@WesleyTran0 

# Feature List

- Added function in classPageInfo to string HTML tags from a string.
- Strips the HTML tag from the course description before putting it into a paragraph tag to be passed into the header body

# Pictures
Before:
image
After:
image

# Reviewers

Primary reviewer:

**Primary**: @mehallhm 

Use the **"Reviewers"** feature in Github

Secondary reviewers: 

**Secondary**: @ItsEricSun @nick-pfeiffer @cherman23

### DIFF
--- a/components/ClassPage/ClassPageInfoBody.tsx
+++ b/components/ClassPage/ClassPageInfoBody.tsx
@@ -18,7 +18,7 @@ export default function ClassPageInfoBody({
       <div className="classPageBodyLeft">
         <HeaderBody
           header="COURSE DESCRIPTION"
-          body={<p>{latestOccurrence.desc}</p>}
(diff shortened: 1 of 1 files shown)


mehallhm:
I would *really* like this in the backend. The reasoning is that these tags are still in the database and we don't want that (makes it hard to search them then!). Ideally the scraper would be updated rather than patching the frontend. We may do both so we don't have to rescrape everything but for now lets keep our focus on the backend.

On that note, this solution is totally viable for a frontend fix!

PR #331, opened 2025-03-24
Prevent notification page from erroring when signing out

cherman23:
# Purpose

Notification page errors when signing out. Fix prevents error by not attempting to access user data. This is a hotfix and the reasoning for the current page reloading when signing out should be investigated. 

# Tickets

no ticket

# Contributors

@cherman23 

# Reviewers

Primary reviewer:

@mehallhm 
@ananyaspatil


mehallhm (approved):
lgtm

